PR #1, opened 2019-01-09
Reinstate initial goroutine check

wdouglass:
Looking at the history of this function, originally the final count was compared against an intiail count instead of 0. should we reinstate the initial count to prevent test cross-contamination?


backkem:
Yea, I guess this would be fine. It can avoid tests failing due to other tests performed earlier leaking routines.

wdouglass:
Yup, only issue now is silently leaking routines. We should probably add the leak checker to more (all?) tests

PR #2, opened 2019-01-23
Repeat CheckRoutines to reduce flakiness

backkem:
Repeat CheckRoutines to reduce flakiness while not blocking
for longer than needed. The routines are now checked every
200ms for a maximum of 10s.


backkem:
I hope sleep actually unblocks the routine. We could do a for select loop but than we would probably need to add it to the `filterRoutines`.

PR #3, opened 2019-02-18
Connection bridge tool

enobufs:
Added connection bridge test used by new sctp and datachannel tests. Typical usage is to connect two SCTP endpoints with this bridge. The bridge allows you to drop, reorder packets as needed.

This relates to pions/webrtc#334

### DIFF
--- a/test/bridge.go
+++ b/test/bridge.go
@@ -1,0 +1,178 @@
+package test
+
+import (
+	"fmt"
+	"net"
+	"sync"
+	"time"
+)
+
+// bridgeConn is a net.Conn that represents an endpoint of the bridge.
+type bridgeConn struct {
+	br     *Bridge
+	id     int
+	readCh chan []byte
+}
+
+// Read reads data, block until the data becomes available.
+func (conn *bridgeConn) Read(b []byte) (int, error) {
+	if data, ok := <-conn.readCh; ok {
+		n := copy(b, data)
+		return n, nil
+	}
+	return 0, fmt.Errorf("bridgeConn closed")
+}
+
+// Write writes data to the bridge.
+func (conn *bridgeConn) Write(b []byte) (int, error) {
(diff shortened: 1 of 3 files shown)


test/bridge.go
```diff
@@ -0,0 +1,196 @@
...
+
+// Reorder inverses the order of packets currently in the specified queue.
+func (br *Bridge) Reorder(fromID int) error {
+	br.mutex.Lock()
+	defer br.mutex.Unlock()
+
+	var err error
+
+	if fromID == 0 {
+		err = inverse(br.queue0to1)
```
Sean-Der: can we just `return inverse(br.queue0to1)` here? I don't think we need to assign this to a var

test/bridge.go
```diff
@@ -0,0 +1,196 @@
...
+// Write writes data to the bridge.
+func (conn *bridgeConn) Write(b []byte) (int, error) {
+	n := len(b)
+	conn.br.Push(b, conn.id)
+	return n, nil
+}
+
+// Close closes the bridge (releases resources used).
+func (conn *bridgeConn) Close() error {
+	close(conn.readCh)
```
Sean-Der: can you add a `isClosed` member and return an error if this is already locked (just so we don't panic at runtime)

test/bridge.go
```diff
@@ -0,0 +1,196 @@
...
+		return n, nil
+	}
+	return 0, fmt.Errorf("bridgeConn closed")
+}
+
+// Write writes data to the bridge.
+func (conn *bridgeConn) Write(b []byte) (int, error) {
+	n := len(b)
+	conn.br.Push(b, conn.id)
+	return n, nil
```
Sean-Der: can you just do `return len(b), nil` just easier as the reader when I don't have to think about purpose of variables

PR #5, opened 2019-03-02
Conn and PacketConn

kixelated:
These are wrappers around net.Conn and net.PacketConn respectively, implementing batched reads/writes when possible.

My only potential change would be to better differentiate the `WriteBatch` methods between both. Perhaps: `ReadPackets`/`WritePackets` and `ReadMessages`/`WriteMessages`

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,6 +36,7 @@ Check out the **[contributing wiki](https://github.com/pions/webrtc/wiki/Contrib
 * [Yutaka Takeda](https://github.com/enobufs) *Bridge, simulates loss and out of order packets*
 * [Sean DuBois](https://github.com/Sean-Der) - *Testing Infra*
 * [Woodrow Douglass](https://github.com/wdouglass) *Fix test flakiness*
+* [Luke Curley](https://github.com/kixelated) *Batched packet support*
 
 ### License
 MIT License - see [LICENSE](LICENSE) for full text
--- a/conn.go
+++ b/conn.go
@@ -1,0 +1,33 @@
+package transport
+
+import (
+	"net"
+)
+
+// Conn is an extension of net.Conn with batched support
+type Conn interface {
+	net.Conn
+
+	// ReadBatch takes a slice of Messages, reading them in bulk if supported.
+	// Returns the number of messages read, which may be less than the full slice.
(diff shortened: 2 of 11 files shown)


Sean-Der:
Sorry for the delay! This is fantastic, so excited.

* Would it be possible to get some test coverage, doesn't have to be extensive but this drops us from ~80 -> ~50

* This seems all specific to wrapping network communication. Do you think we might be able to keep this just in the `internal/ice` package? It would be nice if we could just bubble up 'does this conn support batch` and not have to bleed networking libs into srtp/sctp. 

great work! and sorry to slow this down, I would love to get this merged ASAP

kixelated:
There's three places that take a `net.Conn` or `net.PacketConn` and need batching support: `srtp`, `webrtc/internal/mux`, and `webrtc/internal/ice`. The `packetSingler` code was duplicated in both repositories.

I really like this interface because it takes a generic connection and wraps it for batching support. Either:
1. Wrap the `net.Conn`, reading/writing just a single message.
2. Cast the connection if it already supports batching.
3. Implement the lower level ipv4 package.

`srtp` will use either option 1 or option 2, depending on how the third-party uses the library. We will also be able to implement option 3, either when the library supports TCP or an address can be provided like `WriteRTPTo(*rtp.Packet, net.Addr)`.

`internal/mux` will always use option 2. `internal/ice` will use option 3 for a UDP connection and either option 1 or option 3 for a TCP connection when we add support.

kixelated:
Oh and re: test coverage. It's not a good practice to focus on the percentage. This code is so trivial that unit tests are not useful and just mirror the exact implementation. I can try but don't let coveralls bully you into thinking this is a good development practice.

kixelated:
webrtc with this change: `47 insertions(+), 78 deletions(-)`
srtp with this change: `18 insertions(+), 65 deletions(-)`

kixelated:
sooo.....

sendmmsg actually makes my application slower. The extra cost is building the `ipv4.Message` struct. It doesn't reduce the number of time spend making the syscall apparently, which is a little confusing but hey that's why you benchmark.

Might reopen this in the future if we're really out of bottlenecks and somehow my approach was wrong.

PR #6, opened 2019-03-05
Add module support

kixelated:
foobar

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,0 +1,3 @@
+module github.com/pions/transport
+
+go 1.12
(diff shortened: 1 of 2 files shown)


go.mod
```diff
@@ -0,0 +1,3 @@
+module github.com/pions/transport
+
+go 1.12
```
Sean-Der: Is this a version requirement? We still support older versions I think?

Sean-Der:
`hello world`

go.mod
kixelated: ```The go directive in a go.mod file now indicates the version of the language used by the files within that module. It will be set to the current release (go 1.12) if no existing version is present. If the go directive for a module specifies a version newer than the toolchain in use, the go command will attempt to build the packages regardless, and will note the mismatch only if that build fails.```

PR #7, opened 2019-03-05
Add packetio interface for network buffering

kixelated:
We're not able to use `bytes.Buffer` because it would combine multiple
Writes into a single Read. `packetio.Buffer` provides a similar but is
designed to work with packets, avoiding this situation.

There are similar buffers implemented separarely in webrtc/srtp/ice/mux.
These used channels and were too slow to keep up with the read loop. I
wrote a benchmark for comparison:

```
name                   old time/op    new time/op     delta
BenchmarkBuffer14-8       859ns ± 5%      129ns ± 4%   -84.93%
BenchmarkBuffer140-8      832ns ± 4%      154ns ± 4%   -81.43%
BenchmarkBuffer1400-8     825ns ± 8%      351ns ± 4%   -57.49%

name                   old speed      new speed       delta
BenchmarkBuffer14-8    16.3MB/s ± 5%  107.8MB/s ± 3%  +561.08%
BenchmarkBuffer140-8    168MB/s ± 4%    904MB/s ± 4%  +436.70%
BenchmarkBuffer1400-8  1.70GB/s ± 8%   3.99GB/s ± 4%  +134.74%
```

Note that this implementation has an unbounded buffer, while the channel
implementation has no buffer.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,3 +1,5 @@
 module github.com/pions/transport
 
 go 1.12
+
+require github.com/stretchr/testify v1.3.0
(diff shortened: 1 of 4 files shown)


kixelated:
Not sure if this is the best place for this package; let me know what you guys think.

Sean-Der:
This looks really good, thank you so much for digging into this!

Can you add tests for
* Closing with subs
* Writing with no subs

Then this is a merge from me, we can tag and start getting this out. Nice work :)

PR #8, opened 2019-03-14
Add lossChance to bridge

Sean-Der:
lossChance lets us simulate packet loss using brideConn, this is useful
for testing DTLS. But could also be useful for testing other things in
the future


enobufs:
Once this is merged, I have another branch (sctp-rtx) I'd like to merge.  Are you able to wait for it before tagging?

Issue #13, opened 2019-04-26
NATted evnironment test tool: vnet


enobufs:
## Summary
We need a tool to test ICE / NAT traversal tests.

## Motivation
Testing ICE is HARD. We need a tool to test pion/ice on various types of network configuration.

## Describe alternatives you've considered
We could use [a docker solution](https://github.com/enobufs/nat-traversal-test-using-docker), but the test takes too long and lacks different types of NAT.

## Additional context
Please see [README](https://github.com/pion/transport/blob/feature/vnet/vnet/README.md) under pion/transport/vnet/.

enobufs:
https://github.com/pion/transport/tree/feature/vnet

@hugoArregui @Sean-Der @backkem 
My progress has been slow but the basic framework has been implemented. For pion/ice and pion/stun to be able to switch to vnet, ListenUDP() and Dialer are still missing and I will be implementing those soon, but I would be great to receive early feedback from you guys.

Sean-Der:
That is amazing @enobufs! I am super excited to try this out, reading through it right now.

hugoArregui:
This is amazing! I'm working on the checklist ICE PR, I think we should totally test it with this. Should we collect a list of the different test cases we want to cover so we can help you to develop and test this?

enobufs:
A list of possible test cases will help!
I need to implement STUN server in pion/stun before we start using it... @Sean-Der who do you think the right person for the work would be?

Sean-Der:
We do have [this](https://github.com/pion/stun/blob/master/test/udp_server.go)

enobufs:
Oh nice!! We could the simple STUN server (based on the test code) inside vnet for now.

enobufs:
Hmm, it may create a circular reference... can we add stun/server.go (marked as experimental or something)?

ernado:
About simple STUN server:
Feel free to ctrl+c ctrl+v the https://github.com/gortc/stund if it enough

Also, regarding testing of ICE and STUN:
http://mininet.org/

PR #14, opened 2019-05-20
[WIP] Virtual network - vnet

enobufs:
See #13 for motivations & goals.


Sean-Der:
Fantastic design @enobufs 

I am really happy that it stays out of the default code path (We can just pass this to constructors, but doesn't touch the code bases otherwise)

I will start doing a full review, but it may take me a few days :) at ~5,000 lines it is a good amount for me to grok!

enobufs:
Thanks @Sean-Der !  I totally understand, please take your time.

My idea/goal for vnet is not really to emulate "realistic" (bandwidth/delay) network. It is rather to emulate logical network structure (NAT, interfaces, routing) with simplicity (for automated testing) and minimal impact on normal operation and code base.

More than likely I have done goofy stuff in the code. Please feel free to ask any questions/concerns!

vnet/README.md
```diff
@@ -0,0 +1,234 @@
...
+```
+
+> The instance of `Net` will be the one passed around the project.
+> Net class has public methods for configuration and for application use.
+
+
+## Implementation
+
+### Design Policy
+* Each pion package should have config object which has `Net` (of type vent.Vet) property. (just like how
```
hugoArregui: s/vent.Vet/vnet.Net/

hugoArregui:
@enobufs wondering if we should try to get ride of the net package by re exporting Addr, IP, etc inside vnet. This could be useful to ensure the vnet package is properly spread everywhere

vnet/README.md
```diff
@@ -0,0 +1,234 @@
...
+|net.InterfaceByName()|a.net.InterfaceByName()||
+|net.ListenPacket()|a.net.ListenPacket()||
+|net.ListenUDP()|a.net.ListenUDP()||
+|net.Listen()|a.net.Listen()|(TODO)|
+|net.ListenTCP()|a.net.ListenTCP()||
+|net.Dial()|a.net.Dial()||
+|net.DialUDP()|(not supported)|Use Dial()|
+|net.DialTCP()|(not supported)||
+|net.Interface|vnet.Interface||
+|net.PacketConn|vnet.PacketConn||
```
hugoArregui: it seems the interfaces vnet side is vnet.UDPPacketConn

vnet/net.go
```diff
@@ -0,0 +1,645 @@
...
+func (n *Net) ListenPacket(network string, address string) (net.PacketConn, error) {
+	if n.v == nil {
+		return net.ListenPacket(network, address)
+	}
+
+	return n.v.listenPacket(network, address)
+}
+
+// ListenUDP acts like ListenPacket for UDP networks.
+func (n *Net) ListenUDP(network string, locAddr *net.UDPAddr) (UDPPacketConn, error) {
```
hugoArregui: this should actually return a pointer to the conn to match the net interface: https://golang.org/pkg/net/#ListenUDP

vnet/README.md
enobufs: Oh - that's a horrible typo :p Will fix!

vnet/README.md
enobufs: Ah, yes. There's no vnet.PacketConn introduced.
vnet.UDPPacketConn is an interface to wrap net.UDPConn and vnet.UDPConn actually. I will fix the doc.

vnet/net.go
enobufs: Yeah, that's the thing I have debated.
In order for vnet.ListenUDP() to support both `native` (returns *net.UDPConn) and `virtual` (returns *vnet.UDPConn), we need to introduce an interface that allows both of those pointers. For that reason, vnet.UDPPacketConn had to be introduced.

In many cases, we could get away with using ListenPacket() method and I even thought not supporting vnet.ListenUDP(). (Google encourage users to use ListenPacket() also. But I guess, vnet.ListenUDP() make it a bit easier to migrate to vnet.

vnet/README.md
enobufs: Fixed @ea6c83d

vnet/README.md
enobufs: Updated README at @ea6c83d

vnet/net.go
enobufs: I updated README at @ea6c83d. Does it make sense to you?

enobufs:
> @enobufs wondering if we should try to get ride of the net package by re exporting Addr, IP, etc inside vnet. This could be useful to ensure the vnet package is properly spread everywhere

We could alias all `interface`s in net package, while it is hard to do so with structs because thous would have associated methods and properties (we'd need to implement those in vnet...).  While it is nice to be able to replace all net.Xxx to vnet.Xxx but that would give us much for the effort we'd have to make IMO.  What do you think?

hugoArregui:
I won't go all the way to fully remove net, but I though having alias for things like IP or Addr would present a complete interface and make the code clearer (and also the reviewing!), but then if we have to deal with really specific things we can import net without problem (I don't want to exclude it on the linter or some extreme like that)

vnet/net.go
hugoArregui: it does makes sense! and I can see why ListenPacket will be preferred too, we can probably change that later.

vnet/README.md
```diff
@@ -0,0 +1,243 @@
...
+// Create WAN (a root router).
+wan, err := vnet.NewRouter(&RouterConfig{
+    CIDR:          "0.0.0.0/0",
+    LoggerFactory: logging.NewDefaultLoggerFactory(),
+})
+
+// Create a network.
+// You can specify a static IP for the instance of Net to use. If not specified,
+// router will assign an IP address that is contained in the router's CIDR.
+nw := vnet.NewNet(&vnet.NetworkConfig{
```
hugoArregui: s/NetworkConfig/NetConfig/

Sean-Der:
Just bringing conversation over here (and quoting @enobufs)

```
That was my initial idea, but there are some problems:
* Types / properties cannot be exposed through the interface.
* net.ListenUDP() for instance returns a struct (not interface) and there are associated methods to it. That means, the approach cannot be a drop-in replacement. (net package uses interfaces for conn, addr which is good and we can take advantage of them, but not fully "virtualized" unfortunately, net.UDPConn, net.TCPConn, net.Interface just to name a few.)
* We'd also like to avoid introducing global variables. (so this PR is unavoidable to me either way)
```

My only concern is that when people are debugging issues it will bring them into the `pion/transport` package. I do think we should try and find an alternative (even if it means more code smells) one of the big issues with libwebrtc is how many layers of indirection/big callstacks you can get. 

The nice thing about Pion right now is how quickly people can understand it (and then start hacking on it)

Sean-Der:
@enobufs @hugoArregui 

I do think this is a two way door though! If you want to land in ICE/helps learning we can do that, and then plan on getting it out of the call stack. This is such a massive improvement in how we test (and a great selling point for Pion) 

So many companies struggle with testing their WebRTC implementation. Think of how great it will be if we can say

```
* Pass these options and you can simulate n bandwidth/latency/jitter (no way you can do that today)
* Pass these options to confirm your TURN server is actually operating as expected
```

enobufs:
It's a valid concern and I anticipated we would discuss this at some point.

If we find someone starts hacking transport package, I'd say that is a good sign. He/she may have the same problem to solve and could become a great contributor we would need.

WebRTC is such a complicated beast! Everyone involved in pion has limited time spared for contribution, therefore, we will need to move smart. I am sure webrtc developers are struggling to achieve quality with such a complex system. This reminds me of the author of Winny, a peer-to-peer file sharing system, very complex system, but it was very successful in Japan (a long time ago - around the time Kazaa, Napster, Gnutella was dominating western region). He said, he spent 80% of his time developing testing tools. Anyway, achieving high quality in connectivity is the key to success for pion. Investment in vnet is super reasonable to me for the inevitable complexity. If I see, wIth my user's hat on, pion developers are making an effort to ensure its quality, it is a BIG plus to me.  Let's make pion the best webrtc library!

hugoArregui:
The thing is.. I don't see many ways of accomplishing vnet without transport dependency (or some other dependency), or without touching the prod code paths (at least to branch as this PR does).

The only things I can think of are very dirty dirty tricks that I'm afraid to mention because you won't talk to me again if I did (cough cough preprocessor cough cough). But at that point it doesn't seem worthy. 

This adds a dependency/abstraction but in a way that makes sense, which is something you can say about having multiples packages instead of one repo with everything together, it makes things harder, but there is an underline reason that is worth the price.

Sean-Der:
@hugoArregui PREPROCESSOR IT IS! :)

Maybe we could have `pion/ice` `pion/dtls` `pion/sctp` accept an interface (and `pion/transport` satisfies it?)

My personal goals here
* In the 'base case' vnet should not appear in the call stack. When debugging crashes I want users to be able to work quickly/efficiently.
* In the 'testing case' we should be able to easily pass `vnet` 
* In the 'simulation case' I want developers to be able to simulate their real world networks easily (I should be able to pass `vnet` to multiple `PeerConnection`s and it should just work)

hugoArregui:
Ok, let's iterate over it, this is how it would like the approach @Sean-Der is suggesting:

https://github.com/pion/ice/pull/46/commits/b6abe0266cfb9669f97263cfb8a7598cb3b13a5d 

but can we remove vnet.UDPPacketConn? I don't see how

enobufs:
@Sean-Der 

The current vnet design is pretty close to your goals, but small thing:
> In the 'base case' vnet should not appear in the call stack. When debugging crashes I want users to be able to work quickly/efficiently.

Whether to use vnet or not (a if branch) is currently inside vnet's method like this:
```
// ListenUDP acts like ListenPacket for UDP networks.
func (n *Net) ListenUDP(network string, locAddr *net.UDPAddr) (UDPPacketConn, error) {
	if n.v == nil {
		return net.ListenUDP(network, locAddr)
	}

	return n.v.listenUDP(network, locAddr)
}
```

Therefore, this function call will show up in the call stack.

If we don't want to show this in the stack, then caller (our code) will need to have the similar branch everywhere in our code like this:
```
	if a.net == nil {
		conn, err =  net.ListenUDP(network, locAddr)
	} else {
	        conn, err =  a.net.ListenUDP(network, locAddr)
        }
```

Problem of: our code would get messy VS having one call in the call stack...
Any thoughts?

enobufs:
Also, what net.ListenUDP() returns and a.net.ListenUDP() returns are not the same. vnet already solves that problem by using a new interface vnet.UDPPacketConn so that user code become simple.

enobufs:
Thanks @hugoArregui . You got the point!

hugoArregui:
one extra commit to remove the unneeded vnet extra abstraction: https://github.com/pion/ice/pull/46/commits/530a76cd9d359c5d0dd5d9e55406d87af0eac4ff 

so:

- we still require transport to get the interfaces
- but we won't get transport in a stack trace in prod
- we will still need a bit of overhead in each project to define the interface, but it may be the traceoff we want

wdyt?

enobufs:
+1 for avoiding boilerplate code (the branch) everywhere with an expense of one line in the call stack.

Issue #15, opened 2019-06-23
vnet: Allow listening on 0.0.0.0:x


enobufs:
### Your environment.
  - Version: v0.8.0
  - Browser: n/a

### What did you do?
While adding support of vnet in pion/turn, I noticed the UDPConn listening on 0.0.0.0 did not correctly receive packets from remote.

### What did you expect?
Receive it correctly

### What happened?
The packet was dropped

Support of any-addr (0.0.0.0) was partially implemented but the destination addr comparison logic was not fully implemented.
I have added a new class, udpConnMap which would correctly manage the UDPConn list including the correct comparison against IP addresses including 0.0.0.0.

The issue also includes a minor issue in the lack of default value for NAT binding lifetime for all the tests in pion/turn to pass.

enobufs:
I have also noticed, pion/turnc uses net.DialUDP(). 
I have just added the support of DialUDP() in vnet.

PR #16, opened 2019-06-23
Allow listening on 0.0.0.0:x

enobufs:
See issue #15 for details

I am working on adding vnet support to pion/turn.
With these fixes, I was able to pass a new test using vnet to test communciation between TURN server and client via a virtually hosted NAT.

### DIFF
--- a/vnet/conn_map.go
+++ b/vnet/conn_map.go
@@ -27,10 +27,7 @@ func (m *udpConnMap) insert(conn *UDPConn) error {
 
 		for _, conn := range conns {
 			laddr := conn.LocalAddr().(*net.UDPAddr)
-			if laddr.IP.IsUnspecified() {
-				return fmt.Errorf("address already in use")
-			}
-			if laddr.IP.Equal(udpAddr.IP) {
+			if laddr.IP.IsUnspecified() || laddr.IP.Equal(udpAddr.IP) {
 				return fmt.Errorf("address already in use")
 			}
 		}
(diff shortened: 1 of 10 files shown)


vnet/conn_map.go
```diff
@@ -0,0 +1,135 @@
...
+	}
+}
+
+func (m *udpConnMap) insert(conn *UDPConn) error {
+	udpAddr := conn.LocalAddr().(*net.UDPAddr)
+
+	// check if the port has a listener
+	conns, ok := m.portMap[udpAddr.Port]
+	if ok {
+		if udpAddr.IP.IsUnspecified() {
```
hugoArregui: sorry I don't understand this specific check here, what case are you covering here?

vnet/conn_map.go
```diff
@@ -0,0 +1,135 @@
...
+			return fmt.Errorf("address already in use")
+		}
+
+		for _, conn := range conns {
+			laddr := conn.LocalAddr().(*net.UDPAddr)
+			if laddr.IP.IsUnspecified() {
```
hugoArregui: super nitpick feel free to ignore, I would prefer to put this and the if below together as `if laddr.IP.IsUnspecified() || laddr.IP.Equal(udpAddr.IP)` just because I was trying to understand why `IsUnspecified` will be a problem, then I realised (when reading the if below) that was a normal match

vnet/conn_map.go
enobufs: Say, there is a socket listening on 192.168.0.2:1234.
When you try to listen on 0.0.0.0:1234, it would hit this statement.

When 0.0.0.0 is used as a local address to bind, it tries to bind all IP address your local machine. If there's any IP address already listening to the port, it will end up with 'addrress already in use' error.

vnet/conn_map.go
enobufs: Ok, that might be better!

vnet/conn_map.go
hugoArregui: of course, right, that makes total sense, it's clear now, thanks!

PR #18, opened 2019-06-25
Added ResolveUDPAddr()

enobufs:
See #15 for details

@hugoArregui 
This fix should let pion/ice#46 move forward.

### DIFF
--- a/vnet/net.go
+++ b/vnet/net.go
@@ -6,6 +6,7 @@ import (
 	"math/rand"
 	"net"
 	"strconv"
+	"strings"
 	"sync"
 )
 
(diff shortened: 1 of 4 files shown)


vnet/net.go
```diff
@@ -353,24 +374,21 @@ func (v *vNet) hasIPAddr(ip net.IP) bool {
...
 }
 
+// caller must hold the mutex
 func (v *vNet) allocateLocalAddr(ip net.IP, port int) error {
+	// gather local IP addresses to bind
 	var ips []net.IP
-
-	switch ip.String() {
-	case "0.0.0.0":
+	if ip.IsUnspecified() {
```
hugoArregui: before in the case of "::" getAllIPAddrs will get ipv6, now that has changed, is that a problem?

PR #21, opened 2019-07-03
Allow nill laddr for ListenUDP and DialUDP

### LINKED ISSUE #20: vnet: Allow laddr being nil for ListenUDP and DialUDP
### Your environment.
  - Version: v0.8.2
  - Browser: n/a

### What did you do?
Current pion/ice uses ListenUDP and DialUDP with laddr argument being nil. Added a test using vnet to pion/ice, it panics when laddr is nil.

### What did you expect?
When laddr is nil or laddr.IP is nil, these should be treated as '0.0.0.0:0' or '0.0.0.0' respectively as the corresponding methods in net package do.

### What happened?
Panic from vnet!

enobufs:
Fixes #20

This is to align the behavior of ListenUDP() and DialUDP() methods with that of net package.
Also, improved a debug log message.


enobufs:
Thanks @hugoArregui !

PR #31, opened 2019-08-03
Introduce configurable delay

### LINKED ISSUE #30: vnet: Delay & Jitter options
## Summary
Introduce configurable delay & jitter applied when traversing each router.

## Motivation
To emulate more realistic network behavior.

## Proposal
Add following parameters to RouterConfig
* MinDelay (time.Duration)
* MaxJitter (time.Duration)

The actual delay would be:  minDelay + maxJitter * rand()  // where rand() returns [0, 1).

enobufs:
* Added following parameters to RouterConfig
  - MinDelay
  - MaxJitter
* Added a test for delay feature
* Resolves #30

### DIFF
--- a/vnet/router.go
+++ b/vnet/router.go
@@ -2,6 +2,7 @@ package vnet
 
 import (
 	"fmt"
+	"math/rand"
 	"net"
 	"sync"
 	"sync/atomic"
@@ -37,6 +38,10 @@ type RouterConfig struct {
 	QueueSize int
 	// Effective only when this router has a parent router
 	NATType *NATType
+	// Minimum Delay
+	MinDelay time.Duration
+	// Max Jitter
+	MaxJitter time.Duration
 	// Logger factory
 	LoggerFactory logging.LoggerFactory
 }
@@ -69,6 +74,8 @@ type Router struct {
 	stopFunc      func()                    // requires mutex [x]
 	resolver      *resolver                 // read-only
 	chunkFilters  []ChunkFilter             // requires mutex [x]
+	minDelay      time.Duration             // requires mutex [x]
+	maxJitter     time.Duration             // requires mutex [x]
 	mutex         sync.RWMutex              // thread-safe
 	pushCh        chan struct{}             // writer requires mutex
 	loggerFactory logging.LoggerFactory     // read-only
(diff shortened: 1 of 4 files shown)


vnet/router.go
```diff
@@ -171,14 +180,25 @@ func (r *Router) Start() error {
...
-			select {
-			case <-r.pushCh:
-			case <-cancelCh:
-				break loop
+
+			if d <= 0 {
+				select {
+				case <-r.pushCh:
+				case <-cancelCh:
+					break loop
```
Sean-Der: What does the `break loop` do vs doing nothing at all?

Both times you will fall back into the `for` loop I believe

vnet/router_test.go
```diff
@@ -242,6 +242,81 @@ func TestRouterStandalone(t *testing.T) {
 		assert.Equal(t, int32(1), atomic.LoadInt32(&nCbs1), "should be zero")
 		assert.Equal(t, byte(2), seq, "should be the last chunk")
 	})
+
+	t.Run("Delay with jitter", func(t *testing.T) {
```
Sean-Der: Mind doing smaller tests? Jitter, Delay and Jitter+Delay?

If you don't have the bandwidth I am happy to grab this! Just so people have a smaller test if they need to debug/fix in the future.

vnet/router.go
enobufs: It's one of the confusing syntaxes in Go - and that's how you break out the for-loop, unlike to label for go-to..

* https://play.golang.org/p/HYqlhyNExD3
* https://medium.com/golangspec/labels-in-go-4ffd81932339

vnet/router_test.go
enobufs: You mean, splitting the test I wrote into three?
```
	t.Run("Delay with jitter", func(t *testing.T)
```
I can do it!

enobufs: While attempting to write a test of "Jitter only", I noticed there's a case where the jitter value is not effective. Let me find out another way...

enobufs: @Sean-Der fixed. Let me know how you feel about the changes.

enobufs: By the way, I noticed golang's timer sleeps a bit longer than what is was told.  When timeout duration is set to 20ms, for instance, it wakes up in 20 ~ 24 ms... (a bit longer than I expected)

enobufs:
@Sean-Der If you could review my changes per your feedback, that would be appreciated.

PR #33, opened 2019-08-19
Allow more than one static IP addresses

### LINKED ISSUE #32: vnet: Allow assigning more than one static IP address
## Summary
Currently, vnet allows only one static IP to a Net (=NIC).

## Motivation
We'd like to test cases where we use more than one IP address on your local machine.

## Describe alternatives you've considered
We cannot do this with the real network interface. Vnet is the way to go.

## Additional context
The specific case I would like to use this feature for is to test TURN server which can have separate IP addresses for TURN ports and relay ports.

enobufs:
Resolves #32
See #32 for details.

### DIFF
--- a/vnet/net.go
+++ b/vnet/net.go
@@ -21,7 +21,7 @@ func newMACAddress() net.HardwareAddr {
 
 type vNet struct {
 	interfaces []*Interface // read-only
-	staticIP   net.IP       // read-only
+	staticIPs  []net.IP     // read-only
 	router     *Router      // read-only
 	udpConns   *udpConnMap  // read-only
 	mutex      sync.RWMutex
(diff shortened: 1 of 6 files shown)


enobufs:
I will add more test to increase coverage.

vnet/net.go
```diff
@@ -489,9 +493,21 @@ func NewNet(config *NetConfig) *Net {
 		Flags:        net.FlagUp | net.FlagMulticast,
 	})
 
+	var staticIPs []net.IP
+	for _, ipStr := range config.StaticIPs {
+		if ip := net.ParseIP(ipStr); ip != nil {
```
hugoArregui: should we show a warn in case the ip cannot be parsed?

enobufs: Oh, yes I agree!  I will add a warning log before I merge!

enobufs: Oops...  NewNet() currently does not take a logger. Also, it does not return error either. :(
I believe it should return an error in this case. I am going to file an issue and work on it later as it would be a breaking change.

hugoArregui: perfect! thanks!

enobufs: Filed it as #34.

enobufs:
Thanks @hugoArregui !

PR #35, opened 2019-09-02
Better error message when packetio.Buffer is full

Sean-Der:
Resolves pion/srtp#22


Sean-Der:
Hopefully this is more descriptive. I am also going to upgrade logging so it will give the function/line that messages are being fired from, hopefully will help debugging also.

PR #37, opened 2019-09-15
Fixed NAT binding and filtering logic

### LINKED ISSUE #36: vnet: MappingBehavior parameter has no effect
### Your environment.
  - Version: v0.8.7
  - Browser: n/a

### What did you do?
I am writing a NAT type discovery tool using pion/stun, pion/turn. While adding tests using vnet, I found a bug in vnet.

### What did you expect?
Correctly control NAT mapping behavior from the config param.

### What happened?
NATConfig.MappingBehavior is ignored due to this bug. Filtering Behavior param is incorrectly used as the mapping param.

enobufs:
Fixes #36 

Not only MappingBehavior param was not used, but also the original code also had some mix-up between mapping and filtering keys and there was no test to detect that.

The changes fix above with additional test that confirms that mapping parameter works as expected and filtering behavior issue has been fixed.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -12,6 +12,7 @@ linters:
     - goconst
     - gocyclo
     - dupl
+    - funlen
 
 issues:
   exclude-use-default: false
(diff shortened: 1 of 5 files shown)


enobufs:
@Sean-Der I am making a NAT type discovery tool using pion/turn and pion/stun right now... 
https://github.com/enobufs/go-nats

Sean-Der:
Oh that is super cool! Want to post about in on the Pion's twitter eventually? We also need to write about `vnet` eventually. 

I originally thought moving off of the GitHub wiki would get people more involved (I was wrong). I need to find a nice way we can write blog posts with very little friction.

enobufs:
Yes, +1. Oh...  I have to enable vnet from pion/webrtc...

PR #38, opened 2019-09-22
Support 1:1 NAT behavior

enobufs:
Relates to pion/webrtc#835

Added 1:1 DNAT behavior.
Tested with PR at pion/ice#103

### DIFF
--- a/vnet/nat.go
+++ b/vnet/nat.go
@@ -2,6 +2,7 @@ package vnet
 
 import (
 	"fmt"
+	"net"
 	"sync"
 	"time"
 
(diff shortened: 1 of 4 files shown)


vnet/nat.go
```diff
@@ -89,62 +142,76 @@ func (n *networkAddressTranslator) translateOutbound(from Chunk) (Chunk, error)
...
-				bound:   bound,
-				mapped:  fmt.Sprintf("%s:%d", n.mappedIP, mappedPort),
-				filters: map[string]struct{}{},
-				expires: time.Now().Add(n.natType.MappingLifeTime),
+		if n.natType.Mode == NATModeNAT1To1 {
+			// 1:1 NAT behavior
+			srcAddr := from.SourceAddr().(*net.UDPAddr)
+			srcIP := n.getPairedMappedIP(srcAddr.IP)
+			if srcIP == nil {
+				n.log.Debugf("[%s] drop outbound chunk %s with not route", n.name, from.String())
```
hugoArregui: Wondering if we should default to normal NAT behavior here instead, for example, if you have a host candidate with 1:1 mapping and also a srfxl candidate with something else.

enobufs: Normal (=NAPT) and 1:1 NAT are very different behavior... (Normal has a notion of NAT binding (with finite lifetime) but 1:1 (D)NAT has no such binding. So, I had to introduce NATType.Mode two switch between the "normal" and "1:1 NAT", and Normal is the default behavior.

hugoArregui: Maybe I'm missunderstanding something, but let's say I want to write a case similar to what AWS provides, I want to have a 1:1 NAT for my public->private ip, but I also want to answer to my private ip, maybe I'm misreading something or I don't see the full picture of how the test will be written, but it seems to me the code will try to get the paired mapped ip of the private ip for the second case, and then drop the package. Does it makes sense what I'm saying? Maybe I'm totally missing the point.

PR #46, opened 2020-01-11
Limit the number crypto.rand.Read bytes

at-wat:
As the number of bytes of entropy is limited by 65536
on some js environment.


at-wat:
Node.js looks not having this limitation. Chrome has.

Issue #57, opened 2020-03-03
Move network related internal packages to here


at-wat:
To make `webrtc/internal/mux.Mux`, `transport/vnet.VNet`, and `transport/test.Bridge` compatible with full `net.Conn` features (including `SetDeadline`), it would be better to move `dtls/internal/net/deadline` to this package. Following packages are also better to be moved to here to make package dependencies clean.

- [x] `dtls/internal/net/deadline` (`net.Conn` compatible deadline timer)
- [x] `dtls/internal/net/connctx` (context cancelable `Conn`)
- [x] `udp`
- [x] `dtls/internal/net/dpipe` (datagram like pipe)
- [ ] `webrtc/internal/mux` (packet multiplexer)

stv0g:
This seems to be related to #168 & #204.

@at-wat Can you provide an update which of the points of this issue are missing?
I am willing to work on it :)

stv0g:
I have just migrated `pion/udp` to this repo 🥳

hasheddan:
@stv0g I noticed that in the `pion/udp` migration, [this commit](https://github.com/pion/udp/commit/2589407f52c94a7e1ec63d17df744eabc4fb1ea5) and the corresponding `sync` package were not migrated -- was that intentional?

stv0g:
Yes, that was intentional as I believe the removed functionality is also part of the standard libraries `sync` package.


I was also irritated why the functionality was in this module. It does not seem to be networking related.

hasheddan:
@stv0g I'm not sure introducing the `sync` package was the proper fix for https://github.com/pion/udp/issues/74, but it seems as though after migrating `pion/udp` here we are back to the state in which the data race was observed.

stv0g:
Hi @hasheddan ,

Thanks for the context. I now understand yhe motivation of our custom `sync.WaitGroup` implementation. I think we can add it back. However I would strongly for adding it into `internal/sync`. We should also highlight the difference to thevstandard implementation to avoid the same mistake in the future.

stv0g:
@hasheddan I am going to address this in https://github.com/pion/transport/issues/259

Sean-Der:
I wasn't able to move `webrtc/internal/mux` out because it imports `pion/ice`.

Closing this for now!

PR #58, opened 2020-03-03
Make test.Bridge and packetio.Buffer compatible with net.Conn

at-wat:
Part of #57 
Related to https://github.com/pion/dtls/pull/208

`dtls.Conn` now supports `SetRead/WriteDeadline` if underlying `Conn` supports them.
I would also like to move `webrtc/internal/mux`, which internally use `packetio.Buffer`, to this package and add support of `SetRead/WriteDeadline`.

`deadline` package is temporary duplicated with `dtls` package. I think `pion/transport` is a better place for `deadline` package.


at-wat:
@enobufs please take a look when you have time.

at-wat:
Merging for now to proceed https://github.com/pion/dtls/pull/226.
Please add comments if any.

adwpc:
Great feature! Sorry for late!

PR #61, opened 2020-03-09
Add wrapped ReplayDetector

at-wat:
Add wrapped version of ReplayDetector which is required in SRTP.


at-wat:
Merging for now to proceed https://github.com/pion/srtp/pull/58.
Please add comments if any.

PR #65, opened 2020-04-02
Close bridge.bridgeConn immediately if empty

at-wat:
Instead of waiting the next Tick().

### Reference issue
related to https://github.com/pion/sctp/pull/118


at-wat:
this causes net.Conn incompatibility.

PR #72, opened 2020-06-01
Optimise packetio.Buffer by using a circular buffer

jech:
This more than doubles the performance of the current (highly unrealistic) benchmark, and reduces allocation to virtually nothing.  There are some minor API changes, described in the commit messages.

```
name          old time/op    new time/op    delta
Buffer14-4       552ns ± 8%     176ns ± 2%   -68.10%  (p=0.000 n=8+8)
Buffer140-4      435ns ± 5%     227ns ± 2%   -47.81%  (p=0.000 n=8+8)
Buffer1400-4     929ns ± 3%     385ns ± 2%   -58.55%  (p=0.000 n=8+8)

name          old speed      new speed      delta
Buffer14-4    25.4MB/s ± 8%  79.5MB/s ± 2%  +212.67%  (p=0.000 n=8+8)
Buffer140-4    322MB/s ± 5%   617MB/s ± 2%   +91.49%  (p=0.000 n=8+8)
Buffer1400-4  1.51GB/s ± 3%  3.64GB/s ± 2%  +141.27%  (p=0.000 n=8+8)

name          old alloc/op   new alloc/op   delta
Buffer14-4        138B ± 2%        3B ± 0%   -97.83%  (p=0.000 n=6+8)
Buffer140-4       264B ± 1%        4B ± 0%   -98.48%  (p=0.000 n=7+8)
Buffer1400-4    1.45kB ± 0%    0.01kB ± 0%   -99.52%  (p=0.000 n=8+8)

name          old allocs/op  new allocs/op  delta
Buffer14-4        1.00 ± 0%      0.00       -100.00%  (p=0.000 n=8+8)
Buffer140-4       1.00 ± 0%      0.00       -100.00%  (p=0.000 n=8+8)
Buffer1400-4      1.00 ± 0%      0.00       -100.00%  (p=0.000 n=8+8)
```

### DIFF
--- a/packetio/buffer.go
+++ b/packetio/buffer.go
@@ -110,12 +110,6 @@ func (b *Buffer) Read(packet []byte) (n int, err error) {
 		if len(b.packets) > 0 {
 			first := b.packets[0]
 
-			// This is a packet-based reader/writer so we can't truncate.
-			if len(first) > len(packet) {
-				b.mutex.Unlock()
-				return 0, io.ErrShortBuffer
-			}
-
 			// Remove our packet and continue.
 			b.packets = b.packets[1:]
 			b.size -= len(first)
@@ -124,6 +118,9 @@ func (b *Buffer) Read(packet []byte) (n int, err error) {
 
 			// Actually transfer the data.
 			n := copy(packet, first)
+			if len(first) > len(packet) {
+				return n, io.ErrShortBuffer
+			}
 			return n, nil
 		}
 
--- a/packetio/buffer.go
+++ b/packetio/buffer.go
@@ -2,6 +2,7 @@
 package packetio
 
 import (
+	"errors"
 	"io"
 	"sync"
 	"time"
@@ -9,28 +10,36 @@ import (
 	"github.com/pion/transport/deadline"
 )
 
+var errPacketTooBig = errors.New("packet too big")
+
 // Buffer allows writing packets to an intermediate buffer, which can then be read form.
 // This is verify similar to bytes.Buffer but avoids combining multiple writes into a single read.
 type Buffer struct {
-	mutex   sync.Mutex
-	packets [][]byte
+	mutex sync.Mutex
+
+	// this is a circular buffer.  If head <= tail, then the useful
+	// data is in the interval [head, tail[.  If tail < head, then
+	// the useful data is the union of [head, len[ and [0, tail[.
+	// In order to avoid ambiguity when head = tail, we always leave
+	// an unused byte in the buffer.
+	data       []byte
+	head, tail int
 
 	notify chan struct{}
 	subs   bool
 	closed bool
 
-	// The number of buffered packets in bytes.
-	size int
+	count                 int
+	limitCount, limitSize int
 
-	// The limit on Write in packet count and total size.
-	limitCount int
-	limitSize  int
-
-	// Read deadline timer.
 	readDeadline *deadline.Deadline
 }
 
-// NewBuffer creates a new Buffer object.
+const minSize = 2048
+const cutoffSize = 128 * 1024
+const maxSize = 4 * 1024 * 1024
+
+// NewBuffer creates a new Buffer.
 func NewBuffer() *Buffer {
 	return &Buffer{
 		notify:       make(chan struct{}),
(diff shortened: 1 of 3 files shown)


Sean-Der:
absolutely amazing, nice work @jech 

Mind adding a test that does [AllocsPerRun](https://golang.org/pkg/testing/#AllocsPerRun) just to make sure someone else doesn't regress your work.

After that I can squash and merge!

jech:
Done as you requested.  While I was at it, I added some logic to prevent the allocation from growing beyond what was specified by `SetLimitSize`.

packetio/buffer.go
```diff
@@ -2,88 +2,171 @@
...
-	limitSize  int
-
-	// Read deadline timer.
 	readDeadline *deadline.Deadline
 }
 
-// NewBuffer creates a new Buffer object.
+const minSize = 2048
+const cutoffSize = 128 * 1024
+const maxSize = 4 * 1024 * 1024
```
at-wat, line 40: I think this highly depends on the application and should be configurable from the public API.

jech: Since we grow the buffer exponentially, the exact values have a negligible effect on the execution time.  For example, doubling `minSize` only saves a single allocation.

at-wat: In my understanding, grown buffer size is limited by this `maxSize` at https://github.com/pion/transport/pull/72/files#diff-6415ee44ec80ba409c064c7c902b4183R77-R79.
So, an application requiring more than 4MB of the buffer never work any more. Am I missing something?

jech: In the previous implementation, it was required for an application to set the maximum size with `SetSizeLimit`.  If it didn't, then it would buffer unbounded amounts of data, which would eventually cause it to crash.

In this version, the size of the buffer is limited by `maxSize`.  This prevents the crash even if the application doesn't call `SetSizeLimit`.

This is a temporary workaround.  The proper fix is to pass the size limit as an argument to `NewBuffer`, so that it's impossible to forget to set the buffer size.  That's for later, though, first I want to get this patch in, which improves both efficiency and robustness.

packetio/buffer.go
```diff
@@ -2,88 +2,171 @@
...
+	var newsize int
+	if len(b.data) < cutoffSize {
+		newsize = 2 * len(b.data)
+	} else {
+		newsize = 5 * len(b.data) / 4
+	}
+	if newsize < minSize {
+		newsize = minSize
+	}
+	if newsize > maxSize {
```
at-wat, line 77:
```suggestion
	if b.limitSize <= 0 && newsize > maxSize {
```

jech: If that's the semantics that we want, then we can simplify the code by discarding the `maxSize` constant and initialise `limitSize` to 4MB.  But I prefer `maxSize` to be a hard limit.

Are there any legitimate applications to a buffer larger than 4MB?

Sean-Der:
@jech @at-wat I vote we merge this one. No breaking/behavior changes, and covered with tests!

I am all for simplifying in the future. If you want to cut TODOs and assign them to me and @jech @at-wat I want to make sure we all are in this together :) I just don't want to lose a contributor if we move too slow :(

packetio/buffer.go
at-wat: I'm currently don't have actual application that exceeds 4MB limit.
But the size of the streaming contents today is increasing, this change prohibit to the package to be use in such applications.
I can't understand it should be hard limited to 4MB, user should be able to control that I believe.

at-wat: @jech @Sean-Der I'm sorry for delayed reply. But I think this hard limit potentially break compatibility and limit future possibility of the package.

at-wat:
@Sean-Der feel free to merge it for now. I'll create an another PR to workaround my concern.

jech:
> I'm currently don't have actual application that exceeds 4MB limit.

Please feel free to rework the code to remove the limit, but please do it in a separate commit rather than editing mine.

> But the size of the streaming contents today is increasing, this change prohibit the package to be use in such applications.

Please be aware that the data structure I'm using is not necessarily adapted to very large buffers (on the orders of gigabytes or so).  If you need such huge buffers, you might need to rework the data structurs.

at-wat:
Sorry, I had merged this without checking lint error.
Edited and force pushed two commit messages to fix lint error on master branch.

PR #74, opened 2020-06-20
Add Buffer.ReadContext

josharian:
To allow easier cancellation.
This is duplicative with SetReadDeadline.
The duplication is unfortunate, but contexts
seem like a nicer API as you go further up the stack.


josharian:
This has serious negative performance impact.

PR #116, opened 2020-12-01
(Discussion) Context changes

OrlandoCo:
Context discussion

#### Description

Previos changes has implemented context in read and write methods on Pion, context internally use two methods to check if it's closed `Done` and `Err` these methods internally use a channel and mutex, this imply more contention on hot paths. 

Currently (and maybe Im not fully understanding) there is no need to make the read/write methods cancellable, the read method can be cancelled by `net.Conn` timeout, and write timeouts can be implemented out of Pion internals if required.

The point is by adding more contention on hot paths, the performance degrades a lot, this PR fixed the `Read` method to omit the creation of the noop `context.Background` so we can see the real perfomance:

```
BenchmarkBufferWR14
BenchmarkBufferWR14/Context
BenchmarkBufferWR14/Context-6         	21068410	        55.8 ns/op	 250.84 MB/s
BenchmarkBufferWR14/NoContext
BenchmarkBufferWR14/NoContext-6       	31273467	        39.2 ns/op	 357.45 MB/s
BenchmarkBufferWR140
BenchmarkBufferWR140/Context
BenchmarkBufferWR140/Context-6        	18236595	        65.5 ns/op	2137.44 MB/s
BenchmarkBufferWR140/NoContext
BenchmarkBufferWR140/NoContext-6      	24440551	        48.9 ns/op	2865.81 MB/s
BenchmarkBufferWR1400
BenchmarkBufferWR1400/Context
BenchmarkBufferWR1400/Context-6       	14133746	        84.1 ns/op	16653.53 MB/s
BenchmarkBufferWR1400/NoContext
BenchmarkBufferWR1400/NoContext-6     	17763186	        67.5 ns/op	20729.45 MB/s
BenchmarkBufferWWR14
BenchmarkBufferWWR14/Context
BenchmarkBufferWWR14/Context-6        	20843000	        57.1 ns/op	 245.37 MB/s
BenchmarkBufferWWR14/NoContext
BenchmarkBufferWWR14/NoContext-6      	30034364	        39.8 ns/op	 351.86 MB/s
BenchmarkBufferWWR140
BenchmarkBufferWWR140/Context
BenchmarkBufferWWR140/Context-6       	17966767	        66.3 ns/op	2112.52 MB/s
BenchmarkBufferWWR140/NoContext
BenchmarkBufferWWR140/NoContext-6     	24303013	        49.4 ns/op	2834.95 MB/s
BenchmarkBufferWWR1400
BenchmarkBufferWWR1400/Context
BenchmarkBufferWWR1400/Context-6      	11622256	       101 ns/op	13831.90 MB/s
BenchmarkBufferWWR1400/NoContext
BenchmarkBufferWWR1400/NoContext-6    	14108354	        85.5 ns/op	16366.36 MB/s
```

These can be magnified by all the methods that checks context upstream.

cc: @Sean-Der @tarrencev @at-wat

### DIFF
--- a/packetio/buffer.go
+++ b/packetio/buffer.go
@@ -190,13 +190,19 @@ func (b *Buffer) WriteContext(ctx context.Context, packet []byte) (int, error) {
 // Returns io.EOF if the buffer is closed.
 func (b *Buffer) Read(packet []byte) (n int, err error) {
 	// Return immediately if the deadline is already exceeded.
-	select {
-	case <-b.readDeadline.Done():
-		return 0, &netError{errTimeout, true, true}
-	default:
+	for {
+		ok, notify, n, err := b.tryRead(packet)
+		if ok {
+			return n, err
+		}
+
+		select {
+		case <-b.readDeadline.Done():
+			return 0, &netError{errTimeout, true, true}
+		case <-notify:
+		}
 	}
 
-	return b.ReadContext(context.Background(), packet)
 }
 
 // ReadContext reads from the buffer with context.
--- a/packetio/buffer_test.go
+++ b/packetio/buffer_test.go
@@ -511,22 +511,6 @@ func benchmarkBufferWR(b *testing.B, size int64, write bool, grow int) { // noli
 		}
 	}
 
-	b.Run("NoContext", func(b *testing.B) {
-		b.ResetTimer()
-		b.SetBytes(size)
-
-		for i := 0; i < b.N; i++ {
-			_, err := buffer.Write(packet)
-			if err != nil {
-				b.Fatalf("Write: %v", err)
-			}
-			_, err = buffer.Read(packet)
-			if err != nil {
-				b.Fatalf("Read: %v", err)
-			}
-		}
-		b.StopTimer()
-	})
 	b.Run("Context", func(b *testing.B) {
 		ctx, cancel := context.WithCancel(context.Background())
 		defer cancel()
@@ -540,6 +524,22 @@ func benchmarkBufferWR(b *testing.B, size int64, write bool, grow int) { // noli
 				b.Fatalf("Write: %v", err)
 			}
 			_, err = buffer.ReadContext(ctx, packet)
+			if err != nil {
+				b.Fatalf("Read: %v", err)
+			}
+		}
+		b.StopTimer()
+	})
+	b.Run("NoContext", func(b *testing.B) {
+		b.ResetTimer()
+		b.SetBytes(size)
+
+		for i := 0; i < b.N; i++ {
+			_, err := buffer.Write(packet)
+			if err != nil {
+				b.Fatalf("Write: %v", err)
+			}
+			_, err = buffer.Read(packet)
 			if err != nil {
 				b.Fatalf("Read: %v", err)
 			}


Sean-Der (approved):

Sean-Der:
This looks right to me! If all the tests pass/benchmarks move in a positive direction looks like a great change.

@at-wat has been owning this so he might have a larger plan/idea here

tarrencev:
If I understanding correctly, it seems that the introduction of context into these code paths could significantly degrade performance. While I like context from a ux perspective, perhaps we should further discuss the migration to it in `v3`? It seems like it introduces at least a ~30% regression in performance

Sean-Der:
@tarrencev yea didn't realize what a problem this was going to be!

Would be nice if we could have Write and WriteContext. That seems to be idiomatic Go wise?

packetio/buffer.go
```diff
@@ -190,13 +190,19 @@ func (b *Buffer) WriteContext(ctx context.Context, packet []byte) (int, error) {
 // Returns io.EOF if the buffer is closed.
 func (b *Buffer) Read(packet []byte) (n int, err error) {
 	// Return immediately if the deadline is already exceeded.
-	select {
-	case <-b.readDeadline.Done():
```
at-wat, line 194: Read without deadline check at the beginning lacks compatibility with `net.Conn` as it may return data after deadline exceed.

at-wat: To make the comparison fair, check of `<-ctx.Done()` should also be omitted.

at-wat (commented):
> Currently (and maybe Im not fully understanding) there is no need to make the read/write methods cancellable, the read method can be cancelled by net.Conn timeout, and write timeouts can be implemented out of Pion internals if required.

There is a blocking point in pion/webrtc API (https://github.com/pion/webrtc/pull/1561/files#diff-0c9775993fdc516e2eeb4b3f1c0a17a097a7f01da810f8ed388e74f98de8b79fR22) and it's not easy to make it canceled by SetDeadline.

at-wat:
> These can be magnified by all the methods that checks context upstream.

I think there is no other place that checks context in pion/webrtc, it is basically just inherited to `packetio`. 
I only added context check to the upper layer to https://github.com/pion/webrtc/pull/1561/files#diff-0c9775993fdc516e2eeb4b3f1c0a17a097a7f01da810f8ed388e74f98de8b79fR22, which is only used at the initialization of PeerConnection.

at-wat:
If we remove ctx.Done() check at the beginning of `ReadContext` (it behaves as same as `Read`), performance difference is not significant.

```diff
--- a/packetio/buffer.go
+++ b/packetio/buffer.go
@@ -209,11 +209,6 @@ func (b *Buffer) Read(packet []byte) (n int, err error) {
 // Deadline timer is ignored in this function.
 func (b *Buffer) ReadContext(ctx context.Context, packet []byte) (n int, err error) {
        // Return immediately if the deadline is already exceeded.
-       select {
-       case <-ctx.Done():
-               return 0, &netError{ctx.Err(), true, true}
-       default:
-       }
```

```
BenchmarkBufferWR14/Context-32         	42517759	        30.7 ns/op	 455.38 MB/s
BenchmarkBufferWR14/NoContext-32       	40719651	        30.1 ns/op	 464.83 MB/s
BenchmarkBufferWR140/Context-32        	30142195	        39.7 ns/op	3529.61 MB/s
BenchmarkBufferWR140/NoContext-32      	30648835	        39.0 ns/op	3589.49 MB/s
BenchmarkBufferWR1400/Context-32       	21288296	        55.5 ns/op	25206.67 MB/s
BenchmarkBufferWR1400/NoContext-32     	21987081	        54.5 ns/op	25710.88 MB/s
BenchmarkBufferWWR14/Context-32        	38998137	        30.6 ns/op	 457.52 MB/s
BenchmarkBufferWWR14/NoContext-32      	39662341	        30.2 ns/op	 462.91 MB/s
BenchmarkBufferWWR140/Context-32       	30412392	        39.4 ns/op	3556.90 MB/s
BenchmarkBufferWWR140/NoContext-32     	30944422	        38.6 ns/op	3625.69 MB/s
BenchmarkBufferWWR1400/Context-32      	19318942	        61.9 ns/op	22611.45 MB/s
BenchmarkBufferWWR1400/NoContext-32    	19547092	        60.9 ns/op	23004.53 MB/s
```

OrlandoCo:
> If we remove ctx.Done() check at the beginning of `ReadContext` (it behaves as same as `Read`), performance difference is not significant.
> 
> ```diff
> --- a/packetio/buffer.go
> +++ b/packetio/buffer.go
> @@ -209,11 +209,6 @@ func (b *Buffer) Read(packet []byte) (n int, err error) {
>  // Deadline timer is ignored in this function.
>  func (b *Buffer) ReadContext(ctx context.Context, packet []byte) (n int, err error) {
>         // Return immediately if the deadline is already exceeded.
> -       select {
> -       case <-ctx.Done():
> -               return 0, &netError{ctx.Err(), true, true}
> -       default:
> -       }
> ```
> 
> ```
> BenchmarkBufferWR14/Context-32         	42517759	        30.7 ns/op	 455.38 MB/s
> BenchmarkBufferWR14/NoContext-32       	40719651	        30.1 ns/op	 464.83 MB/s
> BenchmarkBufferWR140/Context-32        	30142195	        39.7 ns/op	3529.61 MB/s
> BenchmarkBufferWR140/NoContext-32      	30648835	        39.0 ns/op	3589.49 MB/s
> BenchmarkBufferWR1400/Context-32       	21288296	        55.5 ns/op	25206.67 MB/s
> BenchmarkBufferWR1400/NoContext-32     	21987081	        54.5 ns/op	25710.88 MB/s
> BenchmarkBufferWWR14/Context-32        	38998137	        30.6 ns/op	 457.52 MB/s
> BenchmarkBufferWWR14/NoContext-32      	39662341	        30.2 ns/op	 462.91 MB/s
> BenchmarkBufferWWR140/Context-32       	30412392	        39.4 ns/op	3556.90 MB/s
> BenchmarkBufferWWR140/NoContext-32     	30944422	        38.6 ns/op	3625.69 MB/s
> BenchmarkBufferWWR1400/Context-32      	19318942	        61.9 ns/op	22611.45 MB/s
> BenchmarkBufferWWR1400/NoContext-32    	19547092	        60.9 ns/op	23004.53 MB/s
> ```

OrlandoCo:
@at-wat If I understand right, the context introduction was just to fix a possible lock here https://github.com/pion/webrtc/pull/1561/files#diff-0c9775993fdc516e2eeb4b3f1c0a17a097a7f01da810f8ed388e74f98de8b79fR22 ?

Sean-Der:
Oh we can fix this another way! @at-wat we should just return early if SRTP isn’t ready, that is what V2 did.

I introduced this behavior change accidentally. Right now things will block between bind + SRTP is reasy, we don’t want that. I just moved things early because I need Bind to fire so the user could get codec errors.

Sean-Der:
Really nice catch, I didn’t understand the impact until now. Will contribute test, then maybe we don’t need context anymore? Or we can do two paths.

at-wat:
> If I understand right, the context introduction was just to fix a possible lock here https://github.com/pion/webrtc/pull/1561/files#diff-0c9775993fdc516e2eeb4b3f1c0a17a097a7f01da810f8ed388e74f98de8b79fR22 ?

@OrlandoCo I think yes. If there is no blocking place in pion/webrtc and pion/srtp code, I'm fine with SetDeadline method. But using context is better to manage lifetime of the things from application side.

OrlandoCo:
> @OrlandoCo I think yes. If there is no blocking place in pion/webrtc and pion/srtp code, I'm fine with SetDeadline method. But using context is better to manage lifetime of the things from application side.

@at-wat I agree on the lifetime, but I think that for webrtc the read/wrtie methods depends on the connection and not the application lifetime, since we have other methods to end a connection like stopping the transceiver or peerconnection.

at-wat:
> Oh we can fix this another way! @at-wat we should just return early if SRTP isn’t ready, that is what V2 did.
>
> I introduced this behavior change accidentally. Right now things will block between bind + SRTP is reasy, we don’t want that. I just moved things early because I need Bind to fire so the user could get codec errors.

@Sean-Der I agree it would be better.

at-wat:
> @at-wat I agree on the lifetime, but I think that for webrtc the read/wrtie methods depends on the connection and not the application lifetime, since we have other methods to end a connection like stopping the transceiver or peerconnection.

@OrlandoCo one main concern of the lifetime for me is a interceptor chain. Several RTP/RTCP read/write will be chained.

OrlandoCo:
> > @at-wat I agree on the lifetime, but I think that for webrtc the read/wrtie methods depends on the connection and not the application lifetime, since we have other methods to end a connection like stopping the transceiver or peerconnection.
> 
> @OrlandoCo one main concern of the lifetime for me is a interceptor chain. Several RTP/RTCP read/write will be chained.

@at-wat context will help to return faster, but without it, on writing/reading return the errors that may terminate the interceptor chain without need to add a channel read on every interceptor, or maybe create a context for the interceptor chain but not for the read/write methods

at-wat:
@OrlandoCo yea, I agree that if there is no blocking point like https://github.com/pion/webrtc/pull/1561/files#diff-0c9775993fdc516e2eeb4b3f1c0a17a097a7f01da810f8ed388e74f98de8b79fR22.

at-wat:
@josharian any ideas? (related: https://github.com/pion/transport/pull/74)

OrlandoCo:
> @OrlandoCo yea, I agree that if there is no blocking point like https://github.com/pion/webrtc/pull/1561/files#diff-0c9775993fdc516e2eeb4b3f1c0a17a097a7f01da810f8ed388e74f98de8b79fR22.

@at-wat Not sure now, but let me take a look @Sean-Der can help us also

Sean-Der:
@OrlandoCo @at-wat This behavior actually existed in `v2` as well. When implementing https://github.com/pion/webrtc/pull/1566 it caused other tests to fail.

If you create a `RTPSender` and then call `Read` it will block until either
* You get a RTCP Packet
* Close the `RTPSender`

at-wat:
I opened https://github.com/pion/transport/pull/117 to make `packetio.Buffer` supports both performance critical usage and usage requiring `net.Conn` compatibility.

PR #117, opened 2020-12-02
Add StrictDeadline option to packetio.Buffer

at-wat:
If StrictDeadline option is set to false (default), Read() may return data even if the deadline is exceeded or context is canceled. It may increases the performance.
Set true to make Buffer behaves more like net.Conn.

This also restore improved test reverted by ac22c334.

Related #116

### DIFF
--- a/packetio/buffer.go
+++ b/packetio/buffer.go
@@ -208,7 +208,7 @@ func (b *Buffer) Write(packet []byte) (int, error) {
 // Blocks until data is available or the buffer is closed.
 // Returns io.ErrShortBuffer is the packet is too small to copy the Write.
 // Returns io.EOF if the buffer is closed.
-func (b *Buffer) Read(packet []byte) (n int, err error) { // nolint:gocognit for performance
+func (b *Buffer) Read(packet []byte) (n int, err error) { //nolint:gocognit
 	if b.strictDeadline {
 		// Return immediately if the deadline is already exceeded.
 		select {
--- a/packetio/buffer.go
+++ b/packetio/buffer.go
@@ -42,6 +42,8 @@ type Buffer struct {
 	limitCount, limitSize int
 
 	readDeadline *deadline.Deadline
+
+	strictDeadline bool
 }
 
 const (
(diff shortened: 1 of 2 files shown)


at-wat:
@Sean-Der @OrlandoCo  this is what https://github.com/pion/transport/pull/116 suggested. It is unrelated to the context introduction.

at-wat:
@Sean-Der @OrlandoCo please take a look.
In webrtc usage, strict deadline handling is not required. This should improve the performance.

This also restores the update of the test reverted by ac22c334b84c0c73ad1806eaf30606bb15aead46

Sean-Der (approved):

Sean-Der:
nice! that makes sense to me, better to not leave the perf on the table :)

Sean-Der:
sorry that I missed this in all the tagging/pushing

jech:
What's the performance gain?  It's difficult to judge without seeing benchmarks if the change is useful.

jech:
Can you give a usage scenario where setting this to false would be useful? Also, why is the default false?

stv0g:
Hi @jech,

@at-wat motivated the change in this comment:

> In webrtc usage, strict deadline handling is not required. This should improve the performance.

stv0g:
And sorry for not CCing you. I did not knew that you wrote this code.

If you strongly disagree with this change, I will revert it.

jech:
> @at-wat motivated the change in this comment:
> 
>     In webrtc usage, strict deadline handling is not required. This should
>     improve the performance.

I don't understand how this can possibly improve performance.  The code
used to drop obsolete packets straight away.  The new default keeps the
obsolete packets around, which means we spend extra time sending obsolete
packets.  Furthermore, it causes extra bufferbloat.

Thus, I would intuitively expect it to decrease performance.

@at-wat, could you please provide benchmarks that show that this change
improves performance?

jech:
> If you strongly disagree with this change, I will revert it.

I think it should be reverted.  At the very least, the default should be
the opposite of what it is now.

stv0g:
Agreed. I will revert the change.

@at-wat Please feel free to reopen if we have some benchmarks :)

PR #118, opened 2020-12-03
Revert adding Context to public API

Sean-Der:
This change caused a ~24% performance decrease

Relates to pion/webrtc#1564

This reverts the following commits
* 26e171fd80cada79f0f2172bd96c2b98662455d8

### DIFF
--- a/packetio/buffer.go
+++ b/packetio/buffer.go
@@ -2,7 +2,6 @@
 package packetio
 
 import (
-	"context"
 	"errors"
 	"io"
 	"sync"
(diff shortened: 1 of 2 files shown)


at-wat:
@Sean-Der I believe this doesn't required to be reverted.
Change in https://github.com/pion/transport/pull/116 is not about the context change but about the deadline handling.

at-wat:
https://github.com/pion/transport/pull/117 adds option to disable strict compatibility to `net.Conn` to speed up. (which is same as https://github.com/pion/transport/pull/116)

OrlandoCo (approved):

OrlandoCo:
> #117 adds option to disable strict compatibility to `net.Conn` to speed up. (which is same as #116)

@at-wat the performance was tested on `peerconnection` without the changes done in transport. Those changes were done because in the benchmakrs no contest was performing worst due to an extra channel check, but it was not intended to make it the default way

at-wat:
@OrlandoCo @Sean-Der I believe packetio.Buffer can support Context version and SetDeadline version without overhead by reverting only https://github.com/pion/transport/pull/117/files#diff-a94716032b0fcd19f6b6379e4f760297d008daa48972e6ba58112857cf8130d8R211-R231 this part.

Sean-Der:
@at-wat But no one is using the Context API at this time right?

Personally I would like to revert it for now, and we can come back. Adding a `Read` and `ReadContext` path in the future is a non-breaking change so we can put it in v3.0.0 but it feels like we are rushing getting context in IMO.

at-wat:
OK, please go a head

PR #121, opened 2020-12-18
Convert packetIO buffer to an interface

OrlandoCo:
#### Description
PacketIO buffer interface will allow to use custom buffer for rtp/rtcp streams

### DIFF
--- a/packetio/buffer.go
+++ b/packetio/buffer.go
@@ -10,11 +10,31 @@ import (
 	"github.com/pion/transport/deadline"
 )
 
+// BufferPacketType allow the buffer to know which packet protocol is writing.
+type BufferPacketType int
+
+const (
+	// RTPBufferPacket indicates the buffer that is handling RTP packets
+	RTPBufferPacket BufferPacketType = 1
+	// RTCPBufferPacket indicates the buffer that is handling RTCP packets
+	RTCPBufferPacket BufferPacketType = 2
+)
+
 var errPacketTooBig = errors.New("packet too big")
 
-// Buffer allows writing packets to an intermediate buffer, which can then be read form.
+// Buffer is an interface to implement a writing/reading packet buffer.
+type Buffer interface {
+	io.ReadWriteCloser
+	Count() int
+	SetLimitCount(limit int)
+	Size() int
+	SetLimitSize(limit int)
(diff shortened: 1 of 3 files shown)


packetio/buffer.go
```diff
@@ -10,11 +10,34 @@ import (
...
+type Buffer interface {
+	Write(packet []byte) (int, error)
+	Read(packet []byte) (int, error)
+	Close() (err error)
+	Count() int
+	SetLimitCount(limit int)
+	Size() int
+	SetLimitSize(limit int)
+	SetReadDeadline(t time.Time) error
+	SetBufferMetadata(t BufferPacketType, ssrc uint32)
```
Sean-Der, line 35: Do we need the metadata in the base class? We may also be handling ICE and DTLS traffic.

packetio/buffer.go
```diff
@@ -10,11 +10,34 @@ import (
...
+	RTPBufferPacket BufferPacketType = 1
+	// RTCPBufferPacket indicates the buffer that is handling RTCP packets
+	RTCPBufferPacket BufferPacketType = 2
+)
+
 var errPacketTooBig = errors.New("packet too big")
 
-// Buffer allows writing packets to an intermediate buffer, which can then be read form.
+// Buffer is an interface to implement a writing/reading packet buffer.
+type Buffer interface {
```
Sean-Der, line 26: Nice! Glad to see this was such an easy change. 

Can/should this just embed stdlib interfaces? Read/Write/Close + Deadline I think might already be something.

packetio/buffer.go
OrlandoCo: I could not find anothwer way to inform the buffer which SSRC and packet type is handling

OrlandoCo: @Sean-Der  You can see what I mean in https://github.com/pion/srtp/pull/126/files

packetio/buffer.go
OrlandoCo: I can set the `io.ReadWriteCloser` but not sure for the other ones

OrlandoCo:
@Sean-Der I removed the metadata from buffer and set stdlib interfaces were possible, to understand the changes please also review https://github.com/pion/srtp/pull/126

PR #132, opened 2021-03-13
Do cleanup when close UDPProxy

winlinvip:
1. Stop all proxy workers when close UDPProxy.
2. Use context to notify and wait for all workers to quit.
3. When context is done, close all UDPConn.

### DIFF
--- a/vnet/udpproxy.go
+++ b/vnet/udpproxy.go
@@ -1,6 +1,7 @@
 package vnet
 
 import (
+	"context"
 	"net"
 	"sync"
 	"time"
@@ -50,7 +51,10 @@ func NewProxy(router *Router) (*UDPProxy, error) {
 
 // Close the proxy, stop all workers.
 func (v *UDPProxy) Close() error {
-	// nolint:godox // TODO: FIXME: Do cleanup.
+	v.workers.Range(func(key, value interface{}) bool {
+		_ = value.(*aUDPProxyWorker).Close()
+		return true
+	})
 	return nil
 }
 
@@ -68,9 +72,14 @@ func (v *UDPProxy) Proxy(client *Net, server *net.UDPAddr) error {
 	worker := &aUDPProxyWorker{
 		router: v.router, mockRealServerAddr: v.mockRealServerAddr,
 	}
+
+	// Create context for cleanup.
+	var ctx context.Context
+	ctx, worker.ctxDisposeCancel = context.WithCancel(context.Background())
+
 	v.workers.Store(server.String(), worker)
 
-	return worker.Proxy(client, server)
+	return worker.Proxy(ctx, client, server)
 }
 
 // A proxy worker for a specified proxy server.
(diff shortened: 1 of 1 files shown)


vnet/udpproxy.go
```diff
@@ -82,6 +86,23 @@ type aUDPProxyWorker struct {
 	// key is vnet client addr, which is net.Addr
 	// value is *net.UDPConn
 	endpoints sync.Map
+
+	// For cleanup.
+	ctxDispose       context.Context
```
at-wat, line 91: https://golang.org/pkg/context/#pkg-overview
Storing Context inside struct is strongly discouraged in the official document.

vnet/udpproxy.go
```diff
@@ -93,13 +114,22 @@ func (v *aUDPProxyWorker) Proxy(client *Net, serverAddr *net.UDPAddr) error { //
 		return err
 	}
 
+	// Create context for cleanup.
+	v.ctxDispose, v.ctxDisposeCancel = context.WithCancel(context.Background())
```
at-wat, line 118: `v.ctxDispose` seems used only in this scope, so it can be a local variable.

vnet/udpproxy.go
```diff
@@ -120,12 +150,21 @@ func (v *aUDPProxyWorker) Proxy(client *Net, serverAddr *net.UDPAddr) error { //
...
+		go func() {
+			<-v.ctxDispose.Done()
+			_ = realSocket.Close()
+		}()
+
 		// Bind address.
 		v.endpoints.Store(addr.String(), realSocket)
 
 		// Got packet from real serverAddr, we should proxy it to vnet.
 		// nolint:godox // TODO: FIXME: Do cleanup.
```
at-wat, line 163: Maybe this TODO comment can be removed?

at-wat (commented):
👍

vnet/udpproxy.go
winlinvip: Yep, I removed it.

vnet/udpproxy.go
winlinvip: It's also used in Close(), which will call `v.ctxDisposeCancel()` to cancel the `v.ctxDispose`.

at-wat: Only `CancelFunc` is used in `Close()`.

vnet/udpproxy.go
winlinvip: Yes, it's the best way to use context, the typical use scenario is pass `ctx context.Context` in the first param, like this:

```
func (v *UDPProxy) Proxy(ctx context.Context, client *Net, server *net.UDPAddr) error

func (v *UDPProxy) Close(ctx context.Context) error
```

But, it will change the interface of `io.Closer`, and the API of vnet which does not use `ctx context.Context` as the first param.  For example, `http.Server` changed the API and introduce a new func `Shutdown`:

```go
func (srv *Server) Close() error
func (srv *Server) Shutdown(ctx context.Context) error
```

There is some use case to keep the context as field of struct, for example:

```
type Request struct {
	ctx context.Context
}

func (r *Request) WithContext(ctx context.Context) *Request
``

So, I think right now, it's ok to use context in the struct, or as the first param of API.

vnet/udpproxy.go
winlinvip: OK, I removed the `ctxDispose` from struct.

at-wat (approved):

winlinvip:
@Sean-Der Should I do more changes about this PR? Or it'll be merged automatically?

Sean-Der:
Hey @winlinvip 

I just added you to the organization. After you accept you can merge the PR yourself (just needs one approval, and you can approve others). Sorry I forgot to add you before :(

winlinvip:
@Sean-Der I accepted the invitation, thanks~

PR #151, opened 2021-09-21
Add Token Bucket Filter for bandwidth limitation

mengelbart:
#### Description

This PR adds an alternative proposal for additional network conditioning to #141 . Instead of adding the network condition to a specific `Net` or `Router`, it adds a filter that embeds a `NIC`. That way, the filter can wrap any `Net` before being passed to a router, to simulate a bandwidth limit on the link from the `Router` to the `Net`.

The filter implemented here is just a very simple form of a Token Bucket Filter, which drops all incoming packets when no tokens are available. A later improvement might add some better queuing behavior.

I think similar filters could be used to implement other kinds of network conditioning like latency and different filters could be combined to provide presets for different environments.

@Antonito @Sean-Der @enobufs What do you think about this approach? I'd be happy to hear your thoughts and feedback on this!

### DIFF
--- a/vnet/tbf.go
+++ b/vnet/tbf.go
@@ -1,0 +1,93 @@
+package vnet
+
+import (
+	"context"
+	"time"
+)
+
+const (
+	// Bit is a single bit
+	Bit = 1
+	// KBit is a kilobit
+	KBit = 1000 * Bit
+	// MBit is a Megabit
+	MBit = 1000 * KBit
+)
+
+// TokenBucketFilter implements a token bucket rate limit algorithm.
+type TokenBucketFilter struct {
+	NIC
+	currentTokensInBucket int
+	c                     chan Chunk
+
+	rate     int
+	maxBurst int
+}
+
+// TBFOption is the option type to configure a TokenBucketFilter
+type TBFOption func(*TokenBucketFilter) error
+
+// TBFRate sets the bitrate of a TokenBucketFilter
+func TBFRate(rate int) TBFOption {
+	return func(t *TokenBucketFilter) error {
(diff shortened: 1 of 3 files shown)


Sean-Der (approved):

vnet/tbf.go
```diff
@@ -0,0 +1,92 @@
...
+
+func (t *TokenBucketFilter) onInboundChunk(c Chunk) {
+	t.c <- c
+}
+
+func (t *TokenBucketFilter) run(ctx context.Context) {
+	// (bitrate * S) / 1000 converted to bytes (divide by 8)
+	// S is the update interval in milliseconds
+	add := (t.rate / 1000) / 8
+	ticker := time.NewTicker(1 * time.Millisecond)
```
Sean-Der, line 74: in `ctx.Done` make sure to close this ticker.

vnet/tbf.go
```diff
@@ -0,0 +1,93 @@
...
+	// KBit is a kilobit
+	KBit = 1000 * Bit
+	// MBit is a Megabit
+	MBit = 1000 * KBit
+)
+
+// TokenBucketFilter implements a token bucket rate limit algorithm.
+type TokenBucketFilter struct {
+	NIC
+	bucket int
```
Sean-Der, line 20: Should this be renamed to `bucketSize`?

Sean-Der (approved):

vnet/tbf.go
mengelbart: I think `bucketSize` could be misleading, because it is the number of tokens currently in the bucket.

PR #152, opened 2021-10-16
Make TBF capacity dynamically configurable

mengelbart:
This PR updates the `TokenBucketFilter` to always use the latest bitrate which was set as an option. This makes it possible to use changing bitrates over time which is necessary to implement some of the tests described in https://www.rfc-editor.org/rfc/rfc8867.html#name-variable-available-capacity.

It changes the Option type to return the reverse option. This makes it a bit easier to temporarily change the bitrate. Since we didn't tag a new version since TBF was added, I guess this should not be a problem.


mengelbart:
I think it is better to re-write the TBF tests before merging this. We don't really need the complete vnet/router setup, just to test the token bucket filter.

mengelbart:
replaced by #155

PR #157, opened 2021-11-03
Feat/delay simulation

mengelbart:
#### Description

Adds a filter to simulate delay.


mengelbart:
draft because it includes a commit from #154 which should be merged first.

PR #163, opened 2021-12-16
Fix deadlock in connctx.WriteContext

jkralik:
#### Reference issue
Fixes https://github.com/pion/dtls/issues/390


jkralik:
@daenney After the merge is it possible to release a new version of [pions/dtls](https://github.com/pion/dtls) with the fix?

at-wat (approved):
LGTM

PR #177, opened 2022-04-23
Make TBF refill only when new chunks arrive

mengelbart:
This avoids continuously updating the token count, even if no new chunks
arrive. It also updates some default values and uses floats to calculate
the tokens, which makes it more accurate.

### DIFF
--- a/vnet/tbf.go
+++ b/vnet/tbf.go
@@ -1,8 +1,11 @@
 package vnet
 
 import (
+	"math"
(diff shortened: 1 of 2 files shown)


stv0g (approved):
LGTM

stv0g:
I've had to disable the TBF tests for WASM as it seems that the JS runtime is performing poorly to achieve the desired rates

PR #182, opened 2022-05-13
Simplify notifying readers

jech:
Consolidate the notify channel and the subs boolean into a single
field: notify being non-nil indicates that we have blocked readers.

I have to say that I'm still not happy with this code: it seems
backwards that we should allocate a new channel for every packet,
we should be able to reuse the same channel.  However, try as I might,
I am unable to do that without introducing a race.


adriancable:
@jech - can you share the 'raison d'etre' behind this PR? What it helps with, improves, etc.

Definitely reusing the same channel would be a big win, if you figure out how to do that race-free.

stv0g (approved):
LGTM

PR #188, opened 2022-05-22
Add speedy xorBytes to pion/transport

adriancable:
The xorBytes function is duplicated in pion/turn and pion/srtp.
This PR adds an improved XorBytes function to pion/transport.
After this PR is merged, I will then create PRs to remove the
duplicated functions from pion/turn and pion/srtp and use the
pion/transport function instead. This will significantly improve
crypto performance.

### DIFF
--- a/utils/xor/xor_amd64.go
+++ b/utils/xor/xor_amd64.go
@@ -1,0 +1,27 @@
+// Copyright 2018 The Go Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package transport
+
+// xorBytes xors the bytes in a and b. The destination should have enough
+// space, otherwise xorBytes will panic. Returns the number of bytes xor'd.
+func XorBytes(dst, a, b []byte) int {
+	n := len(a)
+	if len(b) < n {
+		n = len(b)
+	}
+	if n == 0 {
+		return 0
+	}
+	_ = dst[n-1]
+	xorBytesSSE2(&dst[0], &a[0], &b[0], n) // amd64 must have SSE2
+	return n
+}
+
+func XorWords(dst, a, b []byte) {
+	XorBytes(dst, a, b)
+}
+
+//go:noescape
+func xorBytesSSE2(dst, a, b *byte, n int)
--- a/utils/xor/xor_amd64.s
+++ b/utils/xor/xor_amd64.s
@@ -1,0 +1,54 @@
+// Copyright 2018 The Go Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+#include "textflag.h"
+
+// func xorBytesSSE2(dst, a, b *byte, n int)
+TEXT ·xorBytesSSE2(SB), NOSPLIT, $0
+	MOVQ  dst+0(FP), BX
+	MOVQ  a+8(FP), SI
+	MOVQ  b+16(FP), CX
+	MOVQ  n+24(FP), DX
+	TESTQ $15, DX            // AND 15 & len, if not zero jump to not_aligned.
+	JNZ   not_aligned
+
+aligned:
+	MOVQ $0, AX // position in slices
+
+loop16b:
+	MOVOU (SI)(AX*1), X0   // XOR 16byte forwards.
+	MOVOU (CX)(AX*1), X1
+	PXOR  X1, X0
+	MOVOU X0, (BX)(AX*1)
+	ADDQ  $16, AX
+	CMPQ  DX, AX
+	JNE   loop16b
+	RET
+
+loop_1b:
+	SUBQ  $1, DX           // XOR 1byte backwards.
+	MOVB  (SI)(DX*1), DI
+	MOVB  (CX)(DX*1), AX
+	XORB  AX, DI
+	MOVB  DI, (BX)(DX*1)
+	TESTQ $7, DX           // AND 7 & len, if not zero jump to loop_1b.
+	JNZ   loop_1b
+	CMPQ  DX, $0           // if len is 0, ret.
+	JE    ret
+	TESTQ $15, DX          // AND 15 & len, if zero jump to aligned.
+	JZ    aligned
+
+not_aligned:
+	TESTQ $7, DX           // AND $7 & len, if not zero jump to loop_1b.
+	JNE   loop_1b
+	SUBQ  $8, DX           // XOR 8bytes backwards.
+	MOVQ  (SI)(DX*1), DI
+	MOVQ  (CX)(DX*1), AX
+	XORQ  AX, DI
+	MOVQ  DI, (BX)(DX*1)
+	CMPQ  DX, $16          // if len is greater or equal 16 here, it must be aligned.
+	JGE   aligned
+
+ret:
+	RET
--- a/utils/xor/xor_arm64.go
+++ b/utils/xor/xor_arm64.go
@@ -1,0 +1,29 @@
+// Copyright 2020 The Go Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package transport
+
+// xorBytes xors the bytes in a and b. The destination should have enough
+// space, otherwise xorBytes will panic. Returns the number of bytes xor'd.
+func XorBytes(dst, a, b []byte) int {
+	n := len(a)
+	if len(b) < n {
+		n = len(b)
+	}
+	if n == 0 {
+		return 0
+	}
+	// make sure dst has enough space
+	_ = dst[n-1]
+
+	xorBytesARM64(&dst[0], &a[0], &b[0], n)
+	return n
+}
+
+func XorWords(dst, a, b []byte) {
+	XorBytes(dst, a, b)
+}
+
+//go:noescape
+func xorBytesARM64(dst, a, b *byte, n int)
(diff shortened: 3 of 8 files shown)


adriancable:
@jech - take a look at this. On arches which don't unconditionally support unaligned access, I now check for alignment on dst, src1, src2 and if they are all aligned, XOR word-wise (i.e. fast XOR). The check is almost always unneeded unless you are doing funny things, since allocations will always end up at aligned addresses anyway. But it's there to be safe, and there's no penalty.

jech:
Looks good to me.  I'll test on ARMv7 and ARM64 when I have some time, perhaps also on MIPS if I can still get my old router to boot.

The code is copied from the Go stdlib, perhaps their copyright should be retained, or at least it should be mentioned?

This contains assembler code for PPC and PPC64.  Do we want to merge untested PPC assembler, or do we prefer to remove it?

jech:
Adrian, could you please merge from branch `adriancable-tests` in github.com/jech/transport ?

adriancable:
@jech - I've merged your tests.

> The code is copied from the Go stdlib, perhaps their copyright should be retained, or at least it should be mentioned?

It should be - unless I missed something, all the Go copyright notices were maintained, for the code that was taken from Go stdlib. Can you point me to the file(s) where this isn't right? Thanks!

> This contains assembler code for PPC and PPC64. Do we want to merge untested PPC assembler, or do we prefer to remove it?

I haven't personally tested it, but since this part is from the Go stdlib without any changes, I think it's fine.

> I disagree. If you're sending a network packet, it's quite reasonable to do something like

That's a very fair point.

jech:
What's the purpose of XorWords?  Maybe we should remove it?

adriancable:
Sure, I guess the point of it is simply to avoid the (tiny!) overhead of the extra check in xorBytes if the buffer size is a multiple of the CPU word width.

I don’t think it’s very useful.

jech:
I vote we remove it.

adriancable:
Sure, I can do that.

adriancable:
> If you're lucky, len(header) is a multiple of wordSize, but you cannot count on it.

The RTP header length is always a multiple of 4, which is of course the wordSize on 32-bit ARM. So if the RTP packet is aligned, so is the payload.

jech:
@Sean-Der A managerial decision is required here.  I hereby nominate you as the manager in charge.  CC @davidzhao.

Adrian has decided to do the ambitious thing: he's included assembly versions from the Go sources for AMD64, ARM64, PPC, PPC64 in addition to the reasonably efficient Go version using `unsafe`.  We now have three possibilities:

1. Keep all the code.  This implies keeping the PPC/PPC64 versions, which should be correct since they're copied from the Go sources, but which we cannot keep.
2. Keep only the AMD64 and ARM64 assembler versions.  This we can easily test, but it means we have assembler code in the Pion codebase.
3. Only keep the Go version.  This implies losing about a factor of 4 on ARM64 and AMD64.

Three comments to help you make your decision:
- Adrian has included a fairly comprehensive test, which I find satisfactory, and which makes me confident that (1) or (2) is a reasonable thing to do.
- Should you choose either (1) or (2), it will be possible to add an ARM assembly version, which is important (while ARM64 is widely available nowadays, ARM SOMs and boards tend, in my experience, to be more robust than the available ARM64 hardware, and hence are a great choice for video monitoring systems).
- The code is likely to give a 20% improvement in CTR performance, it will have no influence on GCM.

Sean-Der:
I say merge it! This is fantastic :)

I am personally fine with ASM (even CGo) as long as it puts no burden on the user. If everything is a `go get ...` and as portable as before that is fantastic.

I think the next step is communicate to the Go team that we need their help. They have the testing infrastructure and this is stuff that other people can use. I know we haven't had luck in the past :/ but then this can all be temporary.

Sean-Der (approved):

adriancable:
@jech - FYI, I have written both non-NEON and NEON ARM versions in assembly as well. They both also bring very large performance increases.

The reason I didn't include them yet in this PR is because of the whole issue of knowing whether unaligned accesses are safe on ARM, and/or also having to do a runtime NEON detection check if I wanted to use the NEON version.

I think that probably the right approach for ARM is to assume nothing about alignment or NEON support. I would therefore need to take the (non-NEON) ARM assembly version and modify it to do a fast-copy of the aligned stuff in the middle, and a slow copy of the non-aligned stuff at the beginning and end (which may be none). That way you get almost all the benefits of the fast copy, whether the buffer is aligned or not, without having to detect whether the CPU you are running on at runtime handles unaligned accesses or not.

Does this sound like a plausible approach?

jech:
> The reason I didn't include them yet in this PR is because of the whole issue of knowing whether unaligned accesses are safe on ARM at runtime, and/or also having to do a runtime NEON detection check if I wanted to use the NEON version.

We'd just need to introduce a dependency on x/sys/cpu.  https://pkg.go.dev/golang.org/x/sys/cpu

jech:
> I think the next step is communicate to the Go team that we need their help.

https://github.com/golang/go/issues/53021
https://github.com/golang/go/issues/53023

adriancable:
@jech - x/sys/cpu tells us about NEON but not about aligned vs unaligned access or ARMv5 vs ARMv6 vs ARMv7. There is no usable runtime access to GOARM for example (runtime.goarm is there but not exported) and there have been issues filed against this in the golang repo in the past, but haven't led to any resolution.

Here is what I will think I will make:

- If using ARM and NEON is present, use the NEON assembly version. (no alignment check needed here, because all NEON capable processors support unaligned access)
- If using ARM and NEON is not present, and the dst/src1/src2 are aligned, use the non-NEON assembly version.
- If using ARM, NEON is not present, and dst/src1/src2 are not aligned, use the existing slow version.

I'll add a commit here with that stuff, for your review. If it looks OK then next step, I might make a PR against golang/go. The reason why I say "might" is that PRs against golang/go end up being merged to /dev/null significantly more often than to master. I want to check there is appetite for such a change first before doing any more work on it.

jech:
> x/sys/cpu tells us about NEON but not about aligned vs unaligned access or ARMv5 vs ARMv6 vs ARMv7.

Ah, right.

Just a thought — VFPv3 was introduced with ARMv7.  What about assuming unaligned access if VFPv3 is present?

> Here is what I will think I will make:

Agreed, but please consider doing the VFPv3 check suggested above.

adriancable:
... also because I clearly don't understand Go properly, there's a bug in the alignment check I will fix. (right now it checks the alignment of the slice header, not the alignment of the data in the slice)

adriancable:
@jech - oops, I kept the package name from the go stdlib. I agree that's confusing.

adriancable:
@jech - ugh, I accidentally left an inverted condition in there from testing which meant you wouldn't have actually been running the NEON code. Kindly can you update to https://github.com/pion/transport/pull/188/commits/04906d3156d545a935a4667bf348f58554757730 and then re-run your test? Thanks!

jech:
ARM64 (Cortex-A53, no hardware crypto), updated.  I suspect the time is dominated by the crypto on this CPU.
```
name               old time/op    new time/op    delta
WriteRTP/CTR-100     28.5µs ± 1%    27.3µs ± 1%   -3.90%  (p=0.029 n=4+4)
WriteRTP/CTR-1000     139µs ± 1%     125µs ± 1%   -9.87%  (p=0.029 n=4+4)

name               old speed      new speed      delta
WriteRTP/CTR-100   3.94MB/s ± 0%  4.10MB/s ± 1%   +4.06%  (p=0.029 n=4+4)
WriteRTP/CTR-1000  7.28MB/s ± 1%  8.07MB/s ± 1%  +10.92%  (p=0.029 n=4+4)
```

davidzhao:
Really great work @adriancable! a huge improvement on ARM.

adriancable:
Thanks @davidzhao . Now let's see if we can get some traction on golang/go https://github.com/golang/go/issues/53023 and get this code into the Go stdlib. That will be a huge win for crypto on ARM.

adriancable:
Just created a PR https://github.com/golang/go/pull/53154 which adds the new code to crypto/cipher in the Go stdlib. Let's see where it ends up!

adriancable:
@Sean-Der - are you OK to merge this? (this is a policy question - since I opened the PR, someone else merging may be best practice)

After merging this, I'll then create PRs for pion/turn and pion/srtp to use the new XorBytes in pion/transport.

Sean-Der:
@adriancable Yes! Feel free to merge anything you think is right (with one reviewer) 

@jech are you ok with this PR? He is a much better reviewer/coder then me, so buy in from him is better then me :)

jech:
Yeah, while I haven't checked all the assembly versions myself, I approve:

  - minimal new API (now that it's just one function);
  - the set of tests is reasonably comprehensive;
  - the speedup is noticeable.

I have minor doubts about the package name (xor is too specific, what if we decide to add other operations later?).  There's also a typo in xor_arm, the docstring still mentions xorBytes (with a small x).

jech:
> Regarding the package name, how about just utils?

Too generic.  "byteops"?

adriancable:
@jech - I think it's OK to have multiple packages under utils if needed. I think `xor` is OK for now since we can't agree a better one. I think `byteops` is too specific, especially since the fast XOR is really a word operation. :-)

jech:
@adriancable Excellent.  Let me know you will submit PRs for srtp and stun (in which case please CC me), or whether you want me to do it.

adriancable:
@jech - if you do it, then I can review it. I think that’s a good combo if you have the time. Feel free to post crypto benchmarks in the PR!

PR #204, opened 2022-09-01
Add a new interface transport.Net 

### LINKED ISSUE #34: vnet: NewNet() should return an error
There may be a case where a configuration error is detected by NewNew(), but it currently does not return an error. (this was found during the review for #33)

### LINKED ISSUE #168: vnet: Introduce interface for vnet.Net
## Summary

I propose to add an interface for vnet.Net which dependent packages of `vnet.Net` should use (e.g. pion/ice in my use case).

## Motivation

[cunicu](https://github.com/stv0g/cunicu) is user-space daemon to establish Wireguard peer-to-peer connection via ICE.

cunicu needs to provide custom implementations of `net.{Dial,Listen}{UDP,Packet}` in order to attach eBPF filters to the created sockets which filter STUN and let all other traffic pass to a kernel tunnel interface.

See: https://github.com/stv0g/cunicu/blob/master/pkg/proxy/ebpf.go#L43

This allows cunicu to limit itself to handling ICE and stay out of the data path which remains fully in the kernel.

By introducing an interface for `vnet.Net` users could provide custom implementations for most of the functions in the `net` package.

## Describe alternatives you've considered

We are currently providing a custom implementation of `ice.UDPMux` which is already an interface to do the same.

## Additional context

A `ice.UDPMux` is currently only used for host candidates. We cant use this method for relay or reflexive candidates.
See:  https://github.com/pion/ice/issues/400

Also:
- https://github.com/pion/ice/issues/76
- https://github.com/pion/transport/issues/27

### LINKED ISSUE #207: Redefining `vnet`
@Sean-Der @stv0g @Antonito and anyone who cares about "vnet":
I feel `vnet` package should evolve with a fresh eye. From what it is potentially capable of and how broadly its features/updates could affect other packages or even other projects, it may deserve a dedicated repository with its own versions.

The primary focus of this 'vnet' package was #13, for testing complex NAT traversal (UDP hole punching) scenarios for ICE (for which it worked pretty well, found many bugs!), then people's expectations has grown, current API does not seem to be sufficient enough to match the needs.

I'd really hope there would be more contributors to this feature. Drawing more attention from pion community for a new design may lead us to a better place, with healthier sense of ownership of the feature. So, why don't we start with discussing what new version of `vnet` should look like?

Things to think about:
* Redefine goals of the future 'vnet'
* Dedicated repo for this package
* Package name. Using 'vnet' (virtual network) in production (as @stv0g implied) feels wrong.
* API redefined (better interface with pluggability/extensibility)

Relates to #141, #27, #34, #168

stv0g:
#### Description

With two initial implementations
- `stdnet.Net` the functions from the standard libraries `net` package
- `vnet.Net` the virtual network for testing

This PR closes #34 and clearly separates the virtual network 'vnet.Net' from the code
which is using in production. Furthermore this change allows users to provide their own
`transport.Net` implementation for reasons like the one described in #34.

This PR changes some of the core API for this package, why we might consider bumping it to v2.
I have also prepared associated PRs for pion/ice as well as pion/turn which will be linked here shortly.

#### Reference issue

Closes #34
Closes #207 
Closes #168


stv0g:
@Sean-Der Can you unblock to CI for me? I think you already approved the idea of the PR in #168.

stv0g:
@Sean-Der I fixed the CI warnings. Can you re-trigger the pipeline please?

stv0g:
Okay, yet another try please.
Only a single job (the commit message linter) failed.

I think this is ready to be merged @Sean-Der?

stv0g:
Hi @enobufs, @Sean-Der,

is there anything I can do to help getting this PR getting merged?

Do we need more maintainers for Pion?

vnet/net.go
```diff
@@ -525,153 +517,26 @@ func NewNet(config *NetConfig) *Net {
...
-func (n *Net) getStaticIPs() []net.IP {
-	if n.v == nil {
-		return nil
-	}
-
-	return n.v.staticIPs
-}
-
-// IsVirtual tests if the virtual network is enabled.
-func (n *Net) IsVirtual() bool {
```
enobufs, line 655: Don't we need this any more?

enobufs (commented):
Changes look good to me. (just one question)
Before merging this, I'd like to discuss with you how we want to roll out this change. I personally feel we should bump a major version. We'd also need to think about how to update pion/ice and pion/webrtc.  Any thoughts?

enobufs:
@Sean-Der @at-wat Is there anything I should be aware when we bump a major version of this package?

stv0g:
@enobufs I already drafted a bunch of dependent PRs for other Pion packages which will use the new Net interface.

I will update those PRs once the v2 is released.

vnet/net.go
enobufs: Agreed.

at-wat:
@enobufs If there are other pending breaking changes, it would be nice to include them in v2.0.0.
We can tag v2.0.0-rc.0 after this PR if there might be other changes, and tag v2.0.0 after settled.

net/interface.go
```diff
@@ -1,28 +1,20 @@
-package vnet
+package net
```
at-wat, line 1: It might be better to avoid using package name of stdlib.

enobufs: I think @at-wat has a good point.
I have thought about this for a while... I'd like to avoid `vnet` for the interface package. How about `transport`? It's "mostly" about transport layer, and sort of promoting @stv0g 's "Net" interface to be the part of the primary package of this repo.

stv0g: @enobufs I agree :) I will implement this change

stv0g:
Hi @enobufs,

I did some more work and moved the Net interface to the main package of the transport module.
I also started to extend the interface with support for TCP connections.
My goal is to extend the interface as much as we need to cover all uses of the standard Go `net` package throughout the various Pion packages where it makes sense:

- `Dial`, `DialUDP`, `DialTCP` 
- `Listen`, `ListenPacket`, `ListenTCP`, `ListenUDP`
- `ResolveIPAddr`, `ResolveTCPAddr`, `ResolveUDPAddr`
- more?

stv0g:
More changes:
- Harmonized return types of `{stdnet,vnet}.NewNet()`
- Fix typos
- Fix UT assertions
- Add `Net.InterfaceBy{Name,Index}()`
- Make CI pass

enobufs:
I did not realize earlier that this package had not even reached v1 yet.  I am thinking about this work to be the release candidate as v1. (if necessary, we could create v0 (maintenance branch) if necessary (probably not). @at-wat @Sean-Der let me know if you have any thoughts/concerns.

enobufs (approved):
Great work!  Are you able to rebase to fix the conflict? Other than that it looks good.

at-wat:
> Once this landed on master, I will tag it as v1.0.0-beta to facilitate further work on pion/ice.

v0 may have breaking change without bumping major version according to the semantic versioning.
(I don't have strong opinion about this)

enobufs:
> v0 may have breaking change without bumping major version according to the semantic versioning.

True. We could just bump it to v0.8.0 for now, but this package should have v1 soon.

stv0g:
Thanks @enobufs and @at-wat :)

I've rebased and merged the changes. I expect that we might need some more minor patches to make it work with all the other Pion modules. After that we could think about tagging v1 in my opinion.

PR #209, opened 2022-10-20
bugfix: deadline.Set() memory leak (issue #208)

### LINKED ISSUE #208: Deadline package leaks memory
### Your environment.
  - Version: [v0.13.1](https://github.com/pion/dtls/blob/984d41b9981ad38d3b0b20df93b572d0c50750e7/go.mod#L5)
  - Browser: N/A
  - Other Information: N/A

### What did you do?

I utilize the `pion/dtls` package which imports `pion/transport` for a private project. I noticed some scaling issues with my service and upon running pprof I noticed a lot of garbage collection activity related to the deadline package within this repo. This is due to updating the deadline on a DTLS connection via the `pion/dtls` `SetReadDeadline()` method on every packet (we receive upwards of 50kpps on my service) so as you can imagine we are hitting that function call a lot as means of detecting dead DTLS connections. 
Some output from pprof showing most of the allocated memory and objects come from `deadline.Set()` which uses `time.After()` which as currently implemented doesn't release the underlying timer to the garbage collector until it fires (we use a deadline of about 20 seconds). 

196728683-6c9195ed-d2c9-47a0-9414-4964d9657315

`time.After` 's [documentation](https://pkg.go.dev/time#After) states that `The underlying Timer is not recovered by the garbage collector until the timer fires. If efficiency is a concern, use NewTimer instead and call Timer.Stop if the timer is no longer needed.`

A useful blog post explaining a very similar to problem to the one I hit can be found [here](https://www.arangodb.com/2020/09/a-story-of-a-memory-leak-in-go-how-to-properly-use-time-after/)

I think moving towards using `timer.NewTimer()` directly would be more efficient (though I have currently worked around this via other means).

### What did you expect?

Calling `deadline.Set()` should efficiently release unused memory to the garbage collector.

### What happened?

My service that indirectly imports `pion/transport` spent too much time in garbage collection as a result of calling `deadline.Set()` at a high rate.

floatingstatic:
Signed-off-by: Jeremiah Millay <jmillay@fastly.com>

#### Description
Use `time.Newtimer()` paired with `timer.Stop()` to avoid leaking memory with `time.After()` whose underlying timer will not be released to the garbage collector until it has fired.

#### Reference issue
Fixes #208


jech:
> if !timer.Stop() {
>    <-timer.C
> }

Why is that needed?  I'd just say `timer.Stop()` and be done with it.

floatingstatic:
@jech this is the documented best practice for draining the timer's channel after calling stop. See here:

https://pkg.go.dev/time#Timer.Stop

jech:
> @jech this is the documented best practice for draining the timer's channel
> after calling stop. See here:
> 
> https://pkg.go.dev/time#Timer.Stop

Ah, in case the timer is shared between goroutines.  No objection, then.

floatingstatic:
@jech do you need me to do anything further here? Looks like the CI workflow is stuck pending approval.

jech:
You need approval from one of the Pion members.

Antonito (approved):
Hi @floatingstatic,

looks good to me! Thank you for fixing this!

stv0g (approved):
This looks also good to me 👍

PR #225, opened 2023-01-31
Rename Interface.Addresses() to Interface.Addrs()

stv0g:
In order to align our `transport.Interface` wrapper more closely to Go's `net.Interface` type.


at-wat (approved):
LGTM

PR #236, opened 2023-04-12
Add pion/dtls/internal/net/dpipe package

stv0g:
Related to #57

### DIFF
--- a/dpipe/dpipe.go
+++ b/dpipe/dpipe.go
@@ -1,0 +1,144 @@
+// Package dpipe provides the pipe works like datagram protocol on memory.
+package dpipe
+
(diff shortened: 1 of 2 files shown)


dpipe/dpipe.go
```diff
@@ -0,0 +1,144 @@
+// Package dpipe provides the pipe works like datagram protocol on memory.
```
at-wat, line 1: Please add a note to indicate this package is mainly for testing and not for production

stv0g: Thanks :) Done!

at-wat (approved):

PR #245, opened 2023-05-19
Notify read error if udp.listener.readloop exit

### LINKED ISSUE #238: `transport/udp` doesn't return buffers to pool
Reported originally in https://github.com/pion/udp/issues/94

Would you be interested in opening a PR for this @zllovesuki? Would love to add you to the Pion org and get feedback on what we can improve.

thanks!

cnderrauber:
Notify read error if udp.listener.readloop exit,
also fix #238 

#### Description
If listener.readloop encounter an error and exit, the accept will never return as no connection is created.

#### Reference issue
Fixes #...

### DIFF
--- a/AUTHORS.txt
+++ b/AUTHORS.txt
@@ -6,6 +6,7 @@
 Adrian Cable <adrian.cable@gmail.com>
 Atsushi Watanabe <atsushi.w@ieee.org>
 backkem <mail@backkem.me>
+cnderrauber <zengjie9004@gmail.com>
 Hugo Arregui <hugo.arregui@gmail.com>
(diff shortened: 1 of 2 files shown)


boks1971 (approved):
good find @cnderrauber .

Looks like the bug number you mentioned has 2 and 3 flipped. Must be this I guess - https://github.com/pion/transport/issues/238

stv0g:
Oh no, I was to slow to review this :(

I have a question regarding the change. Why do we need the `sync.Pool` here. It seems like the `readLoop` Goroutine could simply allocate the buffer as a local variable. This would simplify the code a lot.

boks1971:
Yeah great point there @stv0g

PR #246, opened 2023-05-19
Remove sync.Pool from udp.listener.readloop

cnderrauber:
As the read buff will be only allocated once, there is no need to use a pool here.

#### Reference issue
Fixes #...


stv0g (approved):

stv0g:
Thanks @cnderrauber :)

PR #252, opened 2023-07-03
Add support for custom UDP connection identifiers

hasheddan:
#### Description

Adds support for custom UDP connection identifiers. If a ConnIDFn is
provided, the result will be used as the identifier of a connection
rather then the remote address. If a connection already exists for the
remote address, its identifier will be updated to the connection ID. If
the remote address for packets associated with a connection identifier
changes, the remote address of the connection will also be updated.

Adds unit testing to ensure that a custom UDP connection identifier
results in expected behavior. Testing adheres to the following steps:

1. Initiate 5 connections using remote address as identifier.
2. Send message on each connection that results in migration from remote
   address identifier to custom identifier.
3. Spawn 20 clients, each assigned to send one message on one of the
   connections (total of 4 messages on each connection).
4. Spawn 5 servers and ensure that each receives the first initiation
   and set messages, then 4 messages, each originating from a different
   remote address.

#### Reference issue

Part of https://github.com/pion/dtls/issues/256

### DIFF
--- a/AUTHORS.txt
+++ b/AUTHORS.txt
@@ -7,6 +7,7 @@ Adrian Cable <adrian.cable@gmail.com>
 Atsushi Watanabe <atsushi.w@ieee.org>
 backkem <mail@backkem.me>
 cnderrauber <zengjie9004@gmail.com>
+Dan Mangum <georgedanielmangum@gmail.com>
 Hugo Arregui <hugo.arregui@gmail.com>
 Jeremiah Millay <jmillay@fastly.com>
 Jozef Kralik <jojo.lwin@gmail.com>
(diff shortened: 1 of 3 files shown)


hasheddan (commented):
Converting this to draft for now after chat with @Sean-Der -- most of this functionality will be implemented in `pion/dtls`.

stv0g:
I agree with @Sean-Der here. Thats a functionality which is better implememented in pion/dtls in my opinion.

stv0g:
@hasheddan Can we close this PR?

hasheddan:
@stv0g sure 👍🏻

PR #253, opened 2023-07-09
Introduce `netctx` package and `PacketConnCtx`

hasheddan:
#### Description

Moves ConnCtx to the netctx package, which will be expanded to include
additional wrapped network interfaces.

Adds a PacketConnCtx that wraps net.PacketConn.

#### Reference issue

Enables https://github.com/pion/dtls/issues/256


connctx/connctx.go
```diff
@@ -2,6 +2,7 @@
 // SPDX-License-Identifier: MIT
 
 // Package connctx wraps net.Conn using context.Context.
+// Deprecated: use netctx instead.
 package connctx
```
hasheddan, line 6: There are packages that currently depend on this, so we deprecate while introducing the same functionality in `netctx`.

stv0g:
Hi @hasheddan,

can you briefly motivate why you renamed the existing package and deprecated the old name?

I am interested to learn about the differences between the two packages. Are there any API breaking changes in it?

hasheddan (commented):
@stv0g in adding `PacketConnCtx`, I had the option to introduce a new package or add it to the existing `connctx` package. The `connctx` package would be a bit of a misnomer for a `PacketConnCtx`, and introducing a new package such as `packetconnctx` would have the same duplicate naming that `connctx` currently has (i.e. `connctx.ConnCtx` and `packetconnctx.PacketConnCtx`). What we are really doing here is wrapping `net` primitives with context, so I instead introduced a package that matches upstream naming so that we have `net.Conn` / `netctx.ConnCtx` and `net.PacketConn` / `netctx.PacketConnCtx`. The API for `ConnCtx` remains the same, except for instead of `New()` the constructor is `NewConnCtx()`. The `connctx` package is kept around to not break existing consumers, but they will start getting linter warnings encouraging them to switch to `netctx`. We can choose to remove the `connctx` after a sufficient deprecation period if needed.

netctx/packetconnctx.go
```diff
@@ -0,0 +1,163 @@
...
+	ReadFromContext(context.Context, []byte) (int, net.Addr, error)
+}
+
+// WriterTo is an interface for context controlled packet writer.
+type WriterTo interface {
+	WriteToContext(context.Context, []byte, net.Addr) (int, error)
+}
+
+// ReadFromWriterTo is a composite of ReadFromWriterTo.
+type ReadFromWriterTo interface {
```
stv0g, line 26: Is this actually used somewhere?

netctx/packetconnctx.go
```diff
@@ -0,0 +1,163 @@
...
+	ReaderFrom
+	WriterTo
+}
+
+// PacketConnCtx is a wrapper of net.PacketConn using context.Context.
+type PacketConnCtx interface {
```
stv0g, line 32: What do you think about removing the `Ctx` from the naming here? Isnt it implicitly given by the package name?

netctx/packetconnctx.go
```diff
@@ -0,0 +1,163 @@
...
+type packetConnCtx struct {
+	nextConn  net.PacketConn
+	closed    chan struct{}
+	closeOnce sync.Once
+	readMu    sync.Mutex
+	writeMu   sync.Mutex
+}
+
+// NewPacketConn creates a new PacketConnCtx wrapping the given net.PacketConn.
+func NewPacketConn(pconn net.PacketConn) PacketConnCtx {
```
stv0g, line 49: Also in reference to the previous commit. I would find it more harmonized if a `NewXXX` actually creates a `XXX`.
But, I would prefer to to simply rename the `(Packet)?ConnCtx` types to `(Packet)?Conn`.

netctx/packetconnctx.go
```diff
@@ -0,0 +1,163 @@
...
+	return p
+}
+
+func (p *packetConnCtx) ReadFromContext(ctx context.Context, b []byte) (int, net.Addr, error) {
+	p.readMu.Lock()
+	defer p.readMu.Unlock()
+
+	select {
+	case <-p.closed:
+		return 0, nil, io.EOF
```
stv0g, line 63: Isn't a `net.ErrClosed` more appropriate here?

netctx/connctx.go
```diff
@@ -0,0 +1,175 @@
...
+	return c
+}
+
+func (c *connCtx) ReadContext(ctx context.Context, b []byte) (int, error) {
+	c.readMu.Lock()
+	defer c.readMu.Unlock()
+
+	select {
+	case <-c.closed:
+		return 0, io.EOF
```
stv0g, line 71: Isn't a `net.ErrClosed` more appropriate here?

stv0g (changes requested):
Hi @hasheddan,

looks good to me :) I have a few smaller comments.
Ideally, we would also comment all the public methods.
I think here we can largely copy and adjust the respective single line comments from the `net` package.

We dont need to copy the full comments. But a single sentence per method would be nice.

netctx/packetconnctx.go
hasheddan: I'm a fan! Only kept it for consistency with the existing `ConnCtx`, but since we are moving it anyway I think we can use `netctx.Conn` and `netctx.PacketConn` 👍🏻

netctx/packetconnctx.go
hasheddan: It isn't -- it matches the `ReadWriter` in `ConnCtx`, but I don't think we need to include it until it is used 👍🏻

netctx/packetconnctx.go
hasheddan: That seems more reasonable to me 👍🏻

hasheddan (commented):
Thanks @stv0g! Updated based on your suggestions!

stv0g (approved):

hasheddan (commented):
Thanks @stv0g! Just updated with the docstrings as well 👍🏻

PR #256, opened 2023-08-01
feat: add latest sequence number notify to replaydetector

hasheddan:
#### Description

Updates the replay detector interface and implementations to support
indicating whether the accepted packet resulted in an update of the
latest sequence number.

Note: this is not a breaking change as the accept() callback was
previously returning zero values and is now returning one, which can be
ignored if desired. It may cause linter failures in some consumers if
using a strict policy.

#### Reference issue

See https://github.com/pion/dtls/pull/558#discussion_r1280695149 for motivation.

### DIFF
--- a/replaydetector/replaydetector.go
+++ b/replaydetector/replaydetector.go
@@ -8,7 +8,14 @@ package replaydetector
 type ReplayDetector interface {
 	// Check returns true if given sequence number is not replayed.
 	// Call accept() to mark the packet is received properly.
-	Check(seq uint64) (accept func(), ok bool)
+	// The return value of accept() indicates whether the accepted packet is
+	// has the latest observed sequence number.
+	Check(seq uint64) (accept func() bool, ok bool)
+}
+
+// nop is a no-op func that is returned in the case that Check() fails.
+func nop() bool {
+	return false
 }
 
 type slidingWindowDetector struct {
@@ -30,30 +37,33 @@ func New(windowSize uint, maxSeq uint64) ReplayDetector {
 	}
 }
 
-func (d *slidingWindowDetector) Check(seq uint64) (accept func(), ok bool) {
+func (d *slidingWindowDetector) Check(seq uint64) (accept func() bool, ok bool) {
 	if seq > d.maxSeq {
 		// Exceeded upper limit.
-		return func() {}, false
+		return nop, false
 	}
 
 	if seq <= d.latestSeq {
 		if d.latestSeq >= uint64(d.windowSize)+seq {
-			return func() {}, false
+			return nop, false
 		}
 		if d.mask.Bit(uint(d.latestSeq-seq)) != 0 {
 			// The sequence number is duplicated.
-			return func() {}, false
+			return nop, false
 		}
 	}
 
-	return func() {
+	return func() bool {
+		latest := seq == 0
 		if seq > d.latestSeq {
 			// Update the head of the window.
 			d.mask.Lsh(uint(seq - d.latestSeq))
 			d.latestSeq = seq
+			latest = true
 		}
 		diff := (d.latestSeq - seq) % d.maxSeq
 		d.mask.SetBit(uint(diff))
+		return latest
 	}, true
 }
 
@@ -75,10 +85,10 @@ type wrappedSlidingWindowDetector struct {
 	init       bool
 }
 
-func (d *wrappedSlidingWindowDetector) Check(seq uint64) (accept func(), ok bool) {
+func (d *wrappedSlidingWindowDetector) Check(seq uint64) (accept func() bool, ok bool) {
 	if seq > d.maxSeq {
 		// Exceeded upper limit.
-		return func() {}, false
+		return nop, false
 	}
 	if !d.init {
 		if seq != 0 {
@@ -99,21 +109,24 @@ func (d *wrappedSlidingWindowDetector) Check(seq uint64) (accept func(), ok bool
 
 	if diff >= int64(d.windowSize) {
 		// Too old.
-		return func() {}, false
+		return nop, false
 	}
 	if diff >= 0 {
 		if d.mask.Bit(uint(diff)) != 0 {
 			// The sequence number is duplicated.
-			return func() {}, false
+			return nop, false
 		}
 	}
 
-	return func() {
+	return func() bool {
+		latest := false
 		if diff < 0 {
 			// Update the head of the window.
 			d.mask.Lsh(uint(-diff))
 			d.latestSeq = seq
+			latest = true
 		}
 		d.mask.SetBit(uint(d.latestSeq - seq))
+		return latest
 	}, true
 }
--- a/replaydetector/replaydetector.go
+++ b/replaydetector/replaydetector.go
@@ -37,7 +37,7 @@ func New(windowSize uint, maxSeq uint64) ReplayDetector {
 	}
 }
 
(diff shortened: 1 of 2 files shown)


replaydetector/replaydetector_test.go
```diff
@@ -84,74 +102,296 @@ func TestReplayDetector(t *testing.T) {
...
+			if !reflect.DeepEqual(c.expected, out) {
+				t.Errorf("Wrong replay detection result:\nexpected: %v\ngot:      %v",
+					c.expected, out,
+				)
+			}
+		})
+	}
+}
+
+func TestReplayDetectorWrapped(t *testing.T) {
```
edaniels, line 212: Is there room to refactor these two top level tests such that the cases are the same but whether we wrap or not is different? Might save a decent amount of lines from being added. If not, all good!

edaniels (approved):
LGTM. I have just one refactor question on the tests but that doesn't block merging this through.

replaydetector/replaydetector_test.go
hasheddan: @edaniels potentially -- they were structured in that manner prior to my changes here. My reasoning for breaking them out was that when adding the check for whether the `latest` was updated on a given input, the test cases ended up with `expected` / `expectedWrap` (which was already present), `valid` / `validWrap`, and `latest` / `latestWrap`, at which point it started to feel like we were trying to run two separate sets of tests in the same test cases. One of the nice attributes of table-driven tests is being able to visually look at a test case and see what behavior it is intended to exercise, which felt like was getting a bit lost (this was probably exacerbated by the wrapped output defaulting to the non-wrapped output when the wrapped was `nil`). The downside of my approach is increase in lines added (as you mentioned) as well as potential lack of maintainability of the two separate sets of cases. I think I personally lean towards the solution I went with here, but I don't feel too strongly. One middle ground option would be to combine them again, but have something like:

```
map[string]struct {
		windowSize uint
		maxSeq     uint64
		input      []uint64
		expected map[string]struct{
			valid      []bool
			latest     []bool
			expected   []uint64
		}
	}
```

Thoughts?

replaydetector/replaydetector.go
```diff
@@ -8,7 +8,14 @@ package replaydetector
 type ReplayDetector interface {
 	// Check returns true if given sequence number is not replayed.
 	// Call accept() to mark the packet is received properly.
-	Check(seq uint64) (accept func(), ok bool)
+	// The return value of accept() indicates whether the accepted packet is
+	// has the latest observed sequence number.
+	Check(seq uint64) (accept func() bool, ok bool)
```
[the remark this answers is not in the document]
hasheddan: @stv0g see commit message / PR body:

> Note: this is not a breaking change as the accept() callback was
previously returning zero values and is now returning one, which can be
ignored if desired. It may cause linter failures in some consumers if
using a strict policy.

replaydetector/replaydetector_test.go
edaniels: Ah yes that could work. I think what I was going for was something like this where we reuse the same cases but extend them where needed:
```golang
type testCase struct {
	windowSize uint
	maxSeq     uint64
	input      []uint64
	valid      []bool
	latest     []bool
	expected   []uint64
}

const (
	largeSeq = 0x100000000000
	hugeSeq  = 0x1000000000000
)

var commonTestCases = map[string]testCase{
	"Continuous": {
		16, 0x0000FFFFFFFFFFFF,
		[]uint64{0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20},
		[]bool{
			true, true, true, true, true, true, true, true, true, true,
			true, true, true, true, true, true, true, true, true, true,
			true,
		},
		[]bool{
			true, true, true, true, true, true, true, true, true, true,
			true, true, true, true, true, true, true, true, true, true,
			true,
		},
		[]uint64{0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20},
	},
}

func checkInput(t *testing.T, det ReplayDetector, c testCase) {
	var out []uint64
	for i, seq := range c.input {
		accept, ok := det.Check(seq)
		if ok != c.valid[i] {
			t.Errorf("Unexpected validity (%d):\nexpected: %v\ngot: %v", seq, c.valid[i], ok)
		}
		if ok {
			out = append(out, seq)
		}
		if latest := accept(); latest != c.latest[i] {
			t.Errorf("Unexpected sequence latest status (%d):\nexpected: %v\ngot: %v", seq, c.latest[i], latest)
		}
	}
	if !reflect.DeepEqual(c.expected, out) {
		t.Errorf("Wrong replay detection result:\nexpected: %v\ngot:      %v",
			c.expected, out,
		)
	}
}

func TestReplayDetector(t *testing.T) {
	for name, c := range commonTestCases {
		c := c
		t.Run(name, func(t *testing.T) {
			det := New(c.windowSize, c.maxSeq)
			checkInput(t, det, c)
		})
	}
}

func TestReplayDetectorWrapped(t *testing.T) {
	wrappedCases := map[string]testCase{
		"WrapContinuous": {
			64, 0xFFFF,
			[]uint64{0xFFFC, 0xFFFD, 0xFFFE, 0xFFFF, 0x0000, 0x0001, 0x0002, 0x0003},
			[]bool{
				true, true, true, true, true, true, true, true,
			},
			[]bool{
				true, true, true, true, true, true, true, true,
			},
			[]uint64{0xFFFC, 0xFFFD, 0xFFFE, 0xFFFF, 0x0000, 0x0001, 0x0002, 0x0003},
		},
	}
	for name, c := range commonTestCases {
		if _, ok := wrappedCases[name]; ok {
			t.Fatalf("duplicate test: %q", name)
		}
		wrappedCases[name] = c
	}
	for name, c := range wrappedCases {
		c := c
		t.Run(name, func(t *testing.T) {
			det := WithWrap(c.windowSize, c.maxSeq)
			checkInput(t, det, c)
		})
	}
}
```

What do you think of that?

hasheddan: @edaniels ah, the issue with doing as you suggest is that the `valid` and `latest` are not the same for every input across wrapped / not wrapped. This is what led me to split them in the first place. A good example is the following (let's say window of `20`:
- `1` (not wrapped: `valid: true` / `latest: true`, wrapped: `valid: true` / `latest: true`)
- `2` (not wrapped: `valid: true` / `latest: true`, wrapped: `valid: true` / `latest: true`)
- `67` (not wrapped: `valid: true` / `latest: true`, wrapped: `valid: true` / `latest: true`)
- `5` (not wrapped: `valid: false` / `latest: false`, wrapped: `valid: true` / `latest: true`)
- `3` (not wrapped: `valid: false` / `latest: false`, wrapped: `valid: true` / `latest: false`)

edaniels: so when mocking this out, I was looking at which test cases are 100% and it's every single case except for the new names introduced in the wrapped variants. Well rather, all the test cases passed so I assume they are equivalent :P

hasheddan: Ah fair point -- I'll pull those out 👍🏻

hasheddan: @edaniels updated 👍🏻 thanks for the feedback!

edaniels (approved):
LGTM!

PR #260, opened 2023-08-27
Fix data race on listener.connWG

### LINKED ISSUE #259: Add back our custom implemetation of `sync.WaitGroup`
We have been using a custom implementation of `sync.WaitGroup` to fix a data race in the `PacketConn` code used by `pion/dtls`

See:
- https://github.com/pion/udp/issues/74
- https://github.com/pion/transport/issues/57#issuecomment-1694528604

stv0g:
Fixes #259

### DIFF
--- a/internal/sync/waitgroup.go
+++ b/internal/sync/waitgroup.go
@@ -1,0 +1,71 @@
+// SPDX-FileCopyrightText: 2023 The Pion community <https://pion.ly>
+// SPDX-License-Identifier: MIT
+
+// Package sync extends basic synchronization primitives.
+package sync
+
+import (
+	"sync"
+)
+
+// A WaitGroup waits for a collection of goroutines to finish.
+// The main goroutine calls Add to set the number of
+// goroutines to wait for. Then each of the goroutines
+// runs and calls Done when finished. At the same time,
+// Wait can be used to block until all goroutines have finished.
+//
+// WaitGroups in the sync package do not allow adding or
+// subtracting from the counter while another goroutine is
+// waiting, while this one does.
+//
+// A WaitGroup must not be copied after first use.
+//
+// In the terminology of the Go memory model, a call to Done
+type WaitGroup struct {
+	c     int64
+	mutex sync.Mutex
+	cond  *sync.Cond
+}
+
+// NewWaitGroup creates a new WaitGroup.
+func NewWaitGroup() *WaitGroup {
+	wg := &WaitGroup{}
+	wg.cond = sync.NewCond(&wg.mutex)
+	return wg
+}
+
+// Add adds delta, which may be negative, to the WaitGroup counter.
+// If the counter becomes zero, all goroutines blocked on Wait are released.
+// If the counter goes negative, Add panics.
+func (wg *WaitGroup) Add(delta int) {
+	wg.mutex.Lock()
+	defer wg.mutex.Unlock()
+	wg.c += int64(delta)
+	if wg.c < 0 {
+		panic("udp: negative WaitGroup counter") // nolint
+	}
+	wg.cond.Signal()
+}
+
+// Done decrements the WaitGroup counter by one.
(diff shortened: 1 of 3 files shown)


internal/sync/waitgroup.go
```diff
@@ -0,0 +1,71 @@
...
+// runs and calls Done when finished. At the same time,
+// Wait can be used to block until all goroutines have finished.
+//
+// WaitGroups in the sync package do not allow adding or
+// subtracting from the counter while another goroutine is
+// waiting, while this one does.
+//
+// A WaitGroup must not be copied after first use.
+//
+// In the terminology of the Go memory model, a call to Done
```
hasheddan, line 23: I noticed this was truncated when this package was introduced in `pion/udp`. Perhaps we could fix it to match https://cs.opensource.google/go/go/+/refs/tags/go1.21.0:src/sync/waitgroup.go;l=23?
```
// In the terminology of the Go memory model, a call to Done
// “synchronizes before” the return of any Wait call that it unblocks.
```

internal/sync/waitgroup.go
```diff
@@ -0,0 +1,71 @@
...
+
+// A WaitGroup waits for a collection of goroutines to finish.
+// The main goroutine calls Add to set the number of
+// goroutines to wait for. Then each of the goroutines
+// runs and calls Done when finished. At the same time,
+// Wait can be used to block until all goroutines have finished.
+//
+// WaitGroups in the sync package do not allow adding or
+// subtracting from the counter while another goroutine is
+// waiting, while this one does.
```
hasheddan, line 19: I was looking at https://pkg.go.dev/sync#WaitGroup.Add and I don't believe this is strictly accurate (I know you are just migrating this package and didn't initially introduce it). Per https://pkg.go.dev/sync#WaitGroup.Add,
```
Note that calls with a positive delta that occur when the counter is zero must happen before a Wait. Calls with a negative delta, or calls with a positive delta that start when the counter is greater than zero, may happen at any time. Typically this means the calls to Add should execute before the statement creating the goroutine or other event to be waited for.
```

I believe the problem that was being encountered in the initial race condition was the rare case in which the wait group counter went to zero, but then [`Accept()` was called on the listener](https://github.com/pion/udp/blob/d4aece4f6a32af411ee19e87f7051e75dd95ab16/conn.go#L48), and its `Add()` raced with `Wait() in the goroutine that is [started when listening beings](https://github.com/pion/udp/blob/d4aece4f6a32af411ee19e87f7051e75dd95ab16/conn.go#L151).

I think the longer term solution is to consider some slight redesign in the UDP listener, but in the mean time, perhaps we could update this comment to mention that it guards against race conditions when adding to a wait group with counter 0 that is concurrently being waited on. What do you think?

hasheddan (commented):
Thanks @stv0g! Left a few minor comments below.

internal/sync/waitgroup.go
stv0g: Thanks for investigation. I was also sceptical about this comment, but then didnt go into the details as I was not the original author.

Looking again at this, I dont believe it would be a good decission to merge this as is.
In my opinion the race should be fixed in the listener not by patching a commonly understood synchronization primitive.

hasheddan: @stv0g sounds good to me -- are you planning to address in this PR or another?

stv0g: Hey @hasheddan,

Sorry, I am currently too busy to look into this in detail.

Would you have the capacity to take this over?
Or should we go with the old fix instead?

hasheddan: Hey @stv0g! I can take a look, thanks for the ping!

edaniels:
It looks like sync.WaitGroup is in use in transport for connWG and there's no current issues around it being used, so I'll close this issue for now.

PR #261, opened 2023-08-29
Add batch read/write for udp listener

cnderrauber:
In case of high throughput that the sfu serves massive audio tracks, the syscall of every single udp.Write will become the largest consumer of cpu time and reduce the throughput. This PR add batch io option to udp listener so the sfu can use udp mux + batch io to merge the syscall to be more efficient. In [livekit](https://github.com/livekit/livekit) sfu performance test, the udp mux with batch io improves nearly 100% forwarding throughput compared to normal config(no udp mux).

### DIFF
--- a/udp/batchconn.go
+++ b/udp/batchconn.go
@@ -4,6 +4,7 @@
 package udp
 
 import (
+	"io"
 	"net"
 	"runtime"
 	"sync"
@@ -28,6 +29,7 @@ type BatchReader interface {
 type BatchPacketConn interface {
 	BatchWriter
 	BatchReader
+	io.Closer
 }
 
 // BatchConn uses ipv4/v6.NewPacketConn to wrap a net.PacketConn to write/read messages in batch,
@@ -48,7 +50,7 @@ type BatchConn struct {
 	closed atomic.Bool
 }
 
-// NewBatchConn creates a *BatchCon from net.PacketConn with batch configs.
+// NewBatchConn creates a *BatchConn from net.PacketConn with batch configs.
 func NewBatchConn(conn net.PacketConn, batchWriteSize int, batchWriteInterval time.Duration) *BatchConn {
 	bc := &BatchConn{
 		PacketConn:         conn,
@@ -92,6 +94,14 @@ func NewBatchConn(conn net.PacketConn, batchWriteSize int, batchWriteInterval ti
 // Close batchConn and the underlying PacketConn
 func (c *BatchConn) Close() error {
 	c.closed.Store(true)
+	c.batchWriteMutex.Lock()
+	if c.batchWritePos > 0 {
+		_ = c.flush()
+	}
+	c.batchWriteMutex.Unlock()
+	if c.batchConn != nil {
+		return c.batchConn.Close()
+	}
 	return c.PacketConn.Close()
 }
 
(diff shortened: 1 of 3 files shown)


udp/batchconn.go
```diff
@@ -0,0 +1,162 @@
...
+	batchWritePos      int
+	batchWriteLast     time.Time
+
+	batchWriteSize     int
+	batchWriteInterval time.Duration
+
+	closed atomic.Bool
+}
+
+// NewBatchConn creates a *BatchCon from net.PacketConn with batch configs.
```
boks1971, line 51: :nit: `*BatchConn`, one `n` missing

boks1971 (commented):
Excellent stuff @cnderrauber as usual!!! Some minor comments.

Would be good to get more :eyes: on this.

udp/conn.go
```diff
@@ -130,11 +158,22 @@ func (lc *ListenConfig) Listen(network string, laddr *net.UDPAddr) (net.Listener
 		lc.Backlog = defaultListenBacklog
 	}
 
+	if lc.Batch.Enable && (lc.Batch.ReadBatchSize <= 0 || lc.Batch.WriteBatchSize <= 0 || lc.Batch.WriteBatchInterval <= 0) {
```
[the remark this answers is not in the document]
cnderrauber: Good point, changed to enable write/read standalone

udp/batchconn.go
```diff
@@ -0,0 +1,161 @@
...
+		batchWriteSize:     batchWriteSize,
+		batchWriteMessages: make([]ipv4.Message, batchWriteSize),
+	}
+	for i := range bc.batchWriteMessages {
+		bc.batchWriteMessages[i].Buffers = [][]byte{make([]byte, sendMTU)}
+	}
+
+	// batch write only supports linux
+	if runtime.GOOS == "linux" {
+		if pc4 := ipv4.NewPacketConn(conn); pc4 != nil {
```
davidzhao, line 66: if I understand this correctly, this will match the underlying packetConn (so that we wrap ipv4 with v4 similarly for v6?).

would udpmux that needs to listen to both v4 and v6 interfaces create multiple NewBatchConn listeners?

udp/batchconn.go
```diff
@@ -0,0 +1,161 @@
...
+
+// WriteTo write message to an UDPAddr, addr should be nil if it is a connected socket.
+func (c *BatchConn) WriteTo(b []byte, addr net.Addr) (int, error) {
+	if c.batchConn == nil {
+		return c.PacketConn.WriteTo(b, addr)
+	}
+	return c.writeBatch(b, addr)
+}
+
+func (c *BatchConn) writeBatch(buf []byte, raddr net.Addr) (int, error) {
```
davidzhao, line 106: should this be named `enqueueMessage` ? a bit confusing to be named `writeBatch` since it conflicts with the public `WriteBatch`

davidzhao (commented):
lgtm! a couple of questions

udp/batchconn.go
cnderrauber: It is already there, the listen on unspecified address is deprecated by `NewMultiUDPMuxFromPorts` in the previous PR that will enumerate interfaces and create UDPMux for each interface.

udp/batchconn.go
```diff
@@ -0,0 +1,161 @@
...
+			}
+		}()
+	}
+
+	return bc
+}
+
+// Close batchConn and the underlying PacketConn
+func (c *BatchConn) Close() error {
+	c.closed.Store(true)
```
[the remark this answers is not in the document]
cnderrauber: Sounds good, will change it

udp/batchconn.go
cnderrauber: the `BatchConn` doesn't have a public `WriteBatch` method, What do you mean the public `WriteBatch`?

udp/batchconn.go
davidzhao: got it, reading the other PR made this clear. 👍

udp/batchconn.go
davidzhao: I meant the private function `writeBatch`, it's a bit confusing to me and I was suggesting changing it to `enqueueMessage`.

This is because:
* `writeBatch` is enqueuing a single message for writes later
* `WriteBatch` is immediately writing a batch of messages

It's not clear to the reader that why the two functions have completely different functionality, but named the same

davidzhao (approved):
lgtm

PR #265, opened 2023-09-02
Revert commit d490a2c which breaks public API

### LINKED ISSUE #264: Breaking change in pion/transport v2.2.2
### Your environment.
  - Version: v2.2.2
  - Browser: N/A
  - Other Information: N/A

### What did you do?
I updated our dependencies and got the update for pion/transport v2.2.2.

### What did you expect?
To be able to build our application without problems.

### What happened?
Compilation error because of a breaking change in pion/transport v2.2.2.

```
Error: ../../../go/pkg/mod/github.com/pion/srtp/v2@v2.0.16/option.go:37:11: cannot use &nopReplayDetector{} (value of type *nopReplayDetector) as type replaydetector.ReplayDetector in return statement:
	*nopReplayDetector does not implement replaydetector.ReplayDetector (wrong type for Check method)
		have Check(uint64) (func(), bool)
		want Check(seq uint64) (accept func() bool, ok bool)
```

The contract for Go modules is to only make breaking change in major version increases (unless the major version i 0). Since transport was at version 2.2.1 before the release of v2.2.2, you can't do breaking changes and still follow the contract for Go modules. If you want to do a breaking change, you have to release a new major version, v3.0.0.

The commit introducing the breaking change is this one: https://github.com/pion/transport/commit/d490a2cb6efc291411b66766587fb377bd15a5f7

When you fix this, you could retract release v2.2.2 in go.mod if you increase the go version in go.mod to 1.16. The retract directive was added in Go 1.16, so you would have to increase the go version for that to work.

Note that a related bug has been filed in the webrtc repo: https://github.com/pion/webrtc/issues/2544

stv0g:
As motivated in #264, we need to revert the change in commit d490a2c as its breaking the API for users downstream.

Fixes #264

### DIFF
--- a/replaydetector/replaydetector.go
+++ b/replaydetector/replaydetector.go
@@ -8,14 +8,7 @@ package replaydetector
 type ReplayDetector interface {
 	// Check returns true if given sequence number is not replayed.
 	// Call accept() to mark the packet is received properly.
-	// The return value of accept() indicates whether the accepted packet is
-	// has the latest observed sequence number.
-	Check(seq uint64) (accept func() bool, ok bool)
-}
-
-// nop is a no-op func that is returned in the case that Check() fails.
-func nop() bool {
-	return false
+	Check(seq uint64) (accept func(), ok bool)
 }
 
 type slidingWindowDetector struct {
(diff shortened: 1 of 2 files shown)


Sean-Der:
Oof sorry @stv0g we started working on this at the same time :( 

I just merged+tagged. Sorry to waste your time on this

stv0g:
Hi @Sean-Der,

no worries :)

PR #290, opened 2024-04-27
Reuse timer in Deadline

paulwe:
before
```
BenchmarkDeadline/Set-10         	 1535445	       784.3 ns/op	     456 B/op	       6 allocs/op
```
after
```
BenchmarkDeadline/Set-10         	13248633	        79.64 ns/op	       0 B/op	       0 allocs/op
```

### DIFF
--- a/deadline/deadline.go
+++ b/deadline/deadline.go
@@ -11,25 +11,46 @@ import (
 	"time"
 )
 
+type deadlineState uint8
+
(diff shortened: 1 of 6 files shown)


boks1971 (approved):
lgtm! But, would be good to get another pair of 👀 on this.

Amazing improvement in the benchmark!

PR #291, opened 2024-04-28
Reuse notify channel when blocked on read

paulwe:


packetio/buffer.go
```diff
@@ -282,14 +280,17 @@ func (b *Buffer) Close() (err error) {
...
+	b.waiting = false
 	b.closed = true
 
 	b.mutex.Unlock()
 
-	if notify != nil {
-		close(notify)
+	if waiting {
+		select {
+		case b.notify <- struct{}{}:
```
cnderrauber, line 291: It will only wake one reading if there are multiple readings blocked and the other will blocked infinite, I believe pion doesn't have this use case now but it would be safer to keep the behavior since it is a public repo.

paulwe: reading concurrently from a reader doesn't make sense. this isn't a reasonable expectation.

cnderrauber: Agree that is not reasonable for the rtp/rtcp packets, just concern if there was other user use it as a normal buffer since it can be used to store any data.

cnderrauber (approved):

Issue #299, opened 2024-07-20
Regression in v2.2.5 - pion/webrtc simulcast tests fail


sirzooro:
I was trying to bump versions of all pion modules uses by v3 of `pion/webrtc`, and found that simulcast tests in pion/webrtc test suite fails with v2.2.5. They work fine with v2.2.4. Here is part of log for failing tests:

```
  ❌ TestPeerConnection_Simulcast_Probe (0s)
  ❌ TestPeerConnection_Simulcast_Probe/Break_NonSimulcast (0s)
goroutine profile: total 42
4 @ 0x8085988 0x80968a8 0x837c0e1 0x84b6f4c 0x83f4d20 0x80b7c81
#	0x837c0e0	github.com/pion/transport/v2/packetio.(*Buffer).Read+0x140	/go/pkg/mod/github.com/pion/transport/v2@v2.2.5/packetio/buffer.go:265
#	0x84b6f4b	github.com/pion/webrtc/v3/internal/mux.(*Endpoint).Read+0x3b	/go/src/github.com/pion/webrtc/internal/mux/endpoint.go:40
#	0x83f4d1f	github.com/pion/srtp/v2.(*session).start.func1+0xbf		/go/pkg/mod/github.com/pion/srtp/v2@v2.0.20/session.go:144

4 @ 0x8085988 0x80b5324 0x80b5300 0x80be490 0x83d660b 0x83dcdd0 0x84cab7e 0x80b7c81
#	0x80b52ff	sync.runtime_notifyListWait+0x14f				/usr/local/go/src/runtime/sema.go:517
#	0x80be48f	sync.(*Cond).Wait+0x9f						/usr/local/go/src/sync/cond.go:70
#	0x83d660a	github.com/pion/sctp.(*Stream).ReadSCTP+0xca			/go/pkg/mod/github.com/pion/sctp@v1.8.19/stream.go:146
#	0x83dcdcf	github.com/pion/datachannel.(*DataChannel).ReadDataChannel+0x3f	/go/pkg/mod/github.com/pion/datachannel@v1.5.8/datachannel.go:198
#	0x84cab7d	github.com/pion/webrtc/v3.(*DataChannel).readLoop+0xcd		/go/src/github.com/pion/webrtc/datachannel.go:361

2 @ 0x8085988 0x804dee2 0x804dadc 0x83dc82b 0x83dc80a 0x84f8246 0x80b7c81
[...]

PR #304, opened 2024-07-22
Make read notify use sync.Cond instead of chan

edaniels:
We've hit a few cases where using a channel as a notify mechanism is subject to races when we are running multi consumer multi producer scenarios. A sync.Cond is more suitable for this given we can properly broadcast when read deadlines are hit, which this change supports.

I thought we were there with https://github.com/pion/transport/commit/d55a60c7e26d064668001e1d8196d92e69b16b4e but it then revealed #299 where reads can miss an event which is especially pertinent in the simulcast probing test.

Fixes: #299

### DIFF
--- a/packetio/buffer.go
+++ b/packetio/buffer.go
@@ -38,14 +38,15 @@ type Buffer struct {
 	data       []byte
 	head, tail int
 
-	notify  chan struct{}
-	waiting bool
-	closed  bool
+	closed bool
 
 	count                 int
 	limitCount, limitSize int
 
-	readDeadline *deadline.Deadline
+	readDeadline              *deadline.Deadline
+	nextDeadline              chan struct{}
+	readNotifier              *sync.Cond
+	readChannelWatcherRunning sync.WaitGroup
 }
 
 const (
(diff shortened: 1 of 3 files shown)


edaniels:
Note: This appears to cause 3 more allocs per test. I'm not sure that truly matters.

paulwe:
the races should be sorted by the recent set of changes. 

this proposal ads significant complexity and overhead with no clear benefit. composing channel reads is a lot easier than juggling multiple synchronization primitives. to use a cond we would have to either make changes to `deadline.Deadline` or bring that feature into `Buffer` so the deadline, write, and close behaviors could share the cond.

edaniels:
> the races should be sorted by the recent set of changes.
> 
> this proposal ads significant complexity and overhead with no clear benefit. composing channel reads is a lot easier than juggling multiple synchronization primitives. to use a cond we would have to either make changes to `deadline.Deadline` or bring that feature into `Buffer` so the deadline, write, and close behaviors could share the cond.

The race that is fixed is one present in #299 where a read is missed. The recent set of changes fixed the closing case (so, all future reads are satisfied). The same critical section https://github.com/pion/transport/blob/master/packetio/buffer.go#L262-L264 still has an issue with concurrent readers. I'll revise this into something simpler hopefully, but it needs to be addressed.

paulwe:
> The same critical section https://github.com/pion/transport/blob/master/packetio/buffer.go#L262-L264 still has an issue with concurrent readers

concurrent reads work as expected now. if there is more than one goroutine waiting in the read loop one makes progress after each write and when the buffer closes they both exit.

PR #306, opened 2024-07-23
Remove unused waiting field

cnderrauber:
Remove unused waiting field


edaniels (approved):
Whoops my bad; that was sloppy. LGTM

PR #307, opened 2024-07-23
Use sync.Cond for packetio.Buffer

paulwe:


edaniels:
@paulwe after sleeping on it, I think I prefer what we landed on last night, sticking to channels. If needed in the future, it will allow us to use context as well pretty easily which would be harder to slot into this solution.

However based on your comments on https://github.com/pion/transport/pull/304, I’m curious what reasoning has changed to go back in the direction of sync.Cond? Maybe I’m not seeing some other added benefit!

paulwe:
in #304 we were using a combination of cond and channels and using a background goroutine to synchronize them. i was concerned about the complexity and possible fragility of that approach. i don't have a preference for one or the other i just think we should pick one.

i wrote this because you expressed a preference for cond and it occurred to me that we could make it work if we abstracted the timer utility out of the deadline helper.

the channel based version is marginally faster for concurrent r/w benchmarks
```
BenchmarkBuffer14-10         	 8565831	       164.2 ns/op	  85.29 MB/s	       2 B/op	       0 allocs/op
BenchmarkBuffer140-10        	 6928408	       144.8 ns/op	 966.76 MB/s	       3 B/op	       0 allocs/op
BenchmarkBuffer1400-10       	 7962511	       167.9 ns/op	8338.08 MB/s	       2 B/op	       0 allocs/op
```
vs
```
BenchmarkBuffer14-10         	 7835804	       155.0 ns/op	  90.34 MB/s	       0 B/op	       0 allocs/op
BenchmarkBuffer140-10        	 7690047	       153.2 ns/op	 914.00 MB/s	       2 B/op	       0 allocs/op
BenchmarkBuffer1400-10       	 6457164	       182.6 ns/op	7668.64 MB/s	       3 B/op	       0 allocs/op
```

edaniels:
> in #304 we were using a combination of cond and channels and using a background goroutine to synchronize them. i was concerned about the complexity and possible fragility of that approach. i don't have a preference for one or the other i just think we should pick one.
> 
> i wrote this because you expressed a preference for cond and it occurred to me that we could make it work if we abstracted the timer utility out of the deadline helper.
> 
> the channel based version is marginally faster for concurrent r/w benchmarks
> 
> ```
> BenchmarkBuffer14-10         	 8565831	       164.2 ns/op	  85.29 MB/s	       2 B/op	       0 allocs/op
> BenchmarkBuffer140-10        	 6928408	       144.8 ns/op	 966.76 MB/s	       3 B/op	       0 allocs/op
> BenchmarkBuffer1400-10       	 7962511	       167.9 ns/op	8338.08 MB/s	       2 B/op	       0 allocs/op
> ```
> 
> vs
> 
> ```
> BenchmarkBuffer14-10         	 7835804	       155.0 ns/op	  90.34 MB/s	       0 B/op	       0 allocs/op
> BenchmarkBuffer140-10        	 7690047	       153.2 ns/op	 914.00 MB/s	       2 B/op	       0 allocs/op
> BenchmarkBuffer1400-10       	 6457164	       182.6 ns/op	7668.64 MB/s	       3 B/op	       0 allocs/op
> ```

Yeah FWIW your solution is a lot cleaner than my spawning an extra goroutine which felt ugly. If the benchmark was significantly in favor sync.Cond, I'd say we should use it. Thank you for exploring this!

PR #337, opened 2025-05-19
Handle replay after wrapping in replaydetector

anders-avos:
#### Description
Fixes the accept() callback of wrappedSlidingWindowDetector to update mask correctly when the sequence number wraps between seq and latestSeq. Without this fix some packets that should be blocked may be let through around sequence number wrap arounds, as demonstrated by the added test case.


at-wat:
CI seems broken since https://github.com/pion/transport/pull/336

Sean-Der (approved):

PR #338, opened 2025-05-28
Make packetio.TestBufferAlloc stable

at-wat:
Fix unstable testing.AllocsPerRun test of packetio.Buffer
- Do not create Buffer in the measured function to measure only allocs during Read/Write
- Do not use test helpers inside the measured function to avoid the effect from updates in the helpers
- Increase the number of runs to filter the noises

#### Reference issue
- The test wasn't properly implemented and returning wrong result on Go 1.24
  - ref: https://github.com/pion/transport/pull/336

### DIFF
--- a/packetio/buffer_test.go
+++ b/packetio/buffer_test.go
@@ -406,48 +406,104 @@ func TestBufferMisc(t *testing.T) {
 	assert.NoError(err)
 }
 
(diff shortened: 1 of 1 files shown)


Sean-Der:
Thank you for fixing this @at-wat sorry I didn't realize it was broken from the Go bump :/

PR #353, opened 2025-09-25
Add dynamic loss and delay filters with runtime controls

hanguyen-nuro:
Implement DelayFilter and LossFilter for vnet package to enable dynamic network condition simulation during runtime:

DelayFilter features:
* Atomic delay adjustment with SetDelay() method
* Automatic startup with background processing
* Thread-safe delay changes without blocking
* Graceful shutdown with packet draining

LossFilter features:
* Multiple loss handler strategies (Random, Shuffle)
* Runtime loss rate adjustment with SetLossRate()
* Immediate or gradual rate change options
* Deterministic vs random loss patterns.

## Testing
* Unit tests

### DIFF
--- a/vnet/delay_filter.go
+++ b/vnet/delay_filter.go
@@ -4,6 +4,7 @@
 package vnet
 
 import (
+	"context"
 	"sync"
 	"sync/atomic"
 	"time"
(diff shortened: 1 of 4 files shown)


mengelbart (approved):
LGTM, but it seems similar to #350? This one passes most of the linter checks, though.

PR #354, opened 2025-09-27
Include all reads in lock

mengelbart:
This fixes a race condition on concurrent reads/writes on the `t.rate` field.


JoeTurki (approved):
thanks

PR #359, opened 2025-10-18
Fix infinite loop when current time equals deadline

mengelbart:
This might not happen in real-world tests, but in synctest, time does not advance if not all goroutines are blocked. But when the deadline equals the timestamp `now`, the timer will be rescheduled to the same time and end up in an infinite loop.

### DIFF
--- a/vnet/delay_filter.go
+++ b/vnet/delay_filter.go
@@ -118,7 +118,7 @@ func (f *DelayFilter) processReadyPackets(now time.Time) {
 		if next == nil {
 			break
 		}
-		if chunk, ok := next.(timedChunk); ok && (chunk.deadline.Equal(now) || chunk.deadline.Before(now)) {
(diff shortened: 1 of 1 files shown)


philipch07 (approved):
Conceptually this makes sense to me, but I have not tested this in depth.

vnet/delay_filter.go
```diff
@@ -118,8 +118,7 @@ func (f *DelayFilter) processReadyPackets(now time.Time) {
 		if next == nil {
 			break
 		}
-
-		if chunk, ok := next.(timedChunk); ok && chunk.deadline.Before(now) {
+		if chunk, ok := next.(timedChunk); ok && (chunk.deadline.Equal(now) || chunk.deadline.Before(now)) {
```
JoeTurki: What do you think about?

```suggestion
		if chunk, ok := next.(timedChunk); ok && !chunk.deadline.After(now) {
```

PR #351, opened 2025-09-11
[vnet] Fix API compatibility and lint issues for vnet delay and loss filter

hanguyen-nuro:
Restore backward compatibility by adding `DelayFilter.Run()` method and maintaining original `NewLossFilter(nic, chance)` signature while preserving new functionality through `NewLossFilterWithHandler()`.

Resolve all 33 golangci-lint issues:
- Replace dynamic errors with static wrapped errors (err113)
- Add proper error checking in tests (errcheck)
- Reduce cyclomatic complexity by splitting large test functions
- Extract DelayFilter helper methods to reduce cognitive complexity
- Fix formatting, variable naming, and code quality issues

Increase DelayFilter test timing tolerance from 10ms to 200ms to handle both native execution overhead from refactoring and WASM environment timing variability.

PR #360, opened 2025-11-04
Fix Loss filter options

JoTurk:
#### Description
Back story: we're making a tool to test and benchmark SCTP (me and @philipch07 ) I was going through vnet filters to add some options, one of them was to add generic `WithSeed` to all filters so we can have reproducible tests.

I found that the option patterns introduced here #353 are actually not really options patterns, more or so a switch. important context before this PR,  I suggested options pattern: https://github.com/pion/transport/pull/351#issuecomment-3333138192 and I didn't have time to review the PR later in time.
 
This is technically a breaking change but I don't think it affects any user.

I added many tests to make sure this doesn't change the behavior, The API is still not super perfect IMO but it's easier to work with now. Also fixed some comments.

### DIFF
--- a/vnet/loss_filter_test.go
+++ b/vnet/loss_filter_test.go
@@ -320,3 +320,184 @@ func TestLossFilterOptionsPattern(t *testing.T) {
 		assert.Greater(t, packets, received)
 	})
 }
+
+func testShuffleRounding(t *testing.T, chance int, blockSize int, expectedReceived int) {
+	t.Helper()
+	mnic := newMockNIC(t)
+	lossFilter, err := NewLossFilterWithOptions(mnic, chance, WithShuffleLossHandler(blockSize))
+	if !assert.NoError(t, err, "should succeed") {
+		return
+	}
+
+	received := 0
+	mnic.mockOnInboundChunk = func(Chunk) {
+		received++
+	}
+
+	for i := 0; i < 100; i++ {
+		lossFilter.onInboundChunk(&chunkUDP{})
+	}
+
+	assert.Equal(t, expectedReceived, received)
+}
+
+func TestLossFilterShuffleRounding(t *testing.T) {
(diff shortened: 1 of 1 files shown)


JoeTurki:
closed by mistake

vnet/loss_filter_test.go
```diff
@@ -320,3 +320,184 @@ func TestLossFilterOptionsPattern(t *testing.T) {
 		assert.Greater(t, packets, received)
 	})
 }
+
+func testShuffleRounding(t *testing.T, chance int, blockSize int, expectedReceived int) {
```
philipch07: Perhaps this should specify that it confirms the number of chunks received. Maybe we should make a snapshot-esque test to confirm the shuffle as well?

vnet/loss_filter.go
```diff
@@ -143,29 +155,53 @@ func (r *RandomShuffleLossHandler) shouldDrop() bool {
 	return res
 }
 
-// LossFilterOption represents a configuration option for LossFilter creation.
-type LossFilterOption func(nic NIC, chance int) (LossFilterHandler, error)
+func (r *lossHandle) setLossRate(chance int, resetImmediately bool) {
+	if r.shuffleBlockSize > 0 {
+		r.setLossRateShuffle(chance, resetImmediately)
```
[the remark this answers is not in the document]
JoeTurki: No it doesn't need lock, yeah setLossRateShuffle handles it

vnet/loss_filter_test.go
JoeTurki: I added this test to confirm the number of chunks received for several (chance, blockSize) combos (e.g., 1%/100 => 99 received; 49%/10 => 50).
we can't test what you're suggesting until we add WithSeed which i plan to add in a next PR, adding WithSeed is the reason im making this PR.

PR #361, opened 2025-11-04
Add vnet duplication filter

JoTurk:
#### Description
add a filter to duplicate packets.
to test network conditions where packets are duplicated intentionally or by a malformed network device etc

Lore: When I was a kid, I had this POS ISP-issued router that duplicated packets at insane rates. It took me a while to figure out what was going on, but it broke all sorts of network applications and poorly coded games network code. once I finally discovered the cause, it felt like having a superpower xd, being able to uncover a bug that was almost impossible to reproduce under normal conditions.

### DIFF
--- a/vnet/duplication_filter.go
+++ b/vnet/duplication_filter.go
@@ -1,0 +1,381 @@
+// SPDX-FileCopyrightText: 2023 The Pion community <https://pion.ly>
+// SPDX-License-Identifier: MIT
+
+package vnet
+
+import (
+	"errors"
+	"math"
+	"math/rand"
+	"sync"
+	"time"
+)
+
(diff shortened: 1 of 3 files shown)


vnet/duplication_filter.go
```diff
@@ -0,0 +1,381 @@
...
+	}
+
+	if err := validateDuplicationConfig(&cfg); err != nil {
+		return nil, err
+	}
+
+	var rng *rand.Rand
+
+	if cfg.seed == nil {
+		// nolint:gosec // weak rand is intended
```
philipch07: Does the modern rand not have a reproducible option via seeds?

PR #365, opened 2025-11-13
ECN support

penhauer:
#### Description

For supporting ECN, it was needed to change the PacketBuffer so it can contain attributes like ECN or TOS. The new PacketBuffer can store packets with additional variable length attributes.

For the rest of pion, it was needed to add a ReadWithAttributes/WriteWithAttributes method to support reading the said packet "features". Many structs involved in the packet path from the socket to the ReceiveRTP method implement either PacketConn or NetConn. These interfaces however, cannot accommodate the needed ReadWithAttributes method. It is exactly the same problem for having a []byte as a packet, as it prevents further feature development. For the said reasons, Two interfaces were introduced, one for PacketConn and one for NetConn.

The names are not the best names, so any suggestions would be appreciated.


packetio/buffer.go
```diff
@@ -119,12 +121,70 @@ func (b *Buffer) grow() error {
...
+
 // Write appends a copy of the packet data to the buffer.
 // Returns ErrFull if the packet doesn't fit.
 //
 // Note that the packet size is limited to 65536 bytes since v0.11.0 due to the internal data structure.
-func (b *Buffer) Write(packet []byte) (int, error) { //nolint:cyclop
-	if len(packet) >= 0x10000 {
+func (b *Buffer) Write(buff []byte) (n int, err error) { //nolint:cyclop
+	return b.WriteWithAttributes(buff, nil)
+}
```
JoeTurki: we shouldn't change the normal write path, this is a behavior change, and it will break applications that depend on this logic.

packetattributes.go
```diff
@@ -0,0 +1,57 @@
+// SPDX-FileCopyrightText: 2025 The Pion community <https://pion.ly>
+// SPDX-License-Identifier: MIT
+
+package transport
+
+const attrMaxLen = 1
+
+type ECN uint8
```
JoeTurki: It might be better if transport's packet-attributes stayed generic, leaving ECN and other metadata details to higher-level application logic.

Because the way i see it is that ECN is a semantics not mechanism, and transport is just a thin layer / helper library, and it shouldn't care about details.

packetattributes.go
```diff
@@ -0,0 +1,57 @@
...
+type PacketAttributesBuffer interface {
+	// for serializing
+	Marshal() []byte
+
+	// for de-serializing. The bytes will be copied into the returned buffer
+	GetBuffer() []byte
+}
+
+type PacketAttributes struct {
+	buffer [attrMaxLen]byte
```
JoeTurki: I think we should make the attribute size dynamic and add a byte to determine how many attributes are there so we don't have to make attributesv2 in the future.

connsocket.go
```diff
@@ -0,0 +1,95 @@
...
+
+import (
+	"net"
+	"time"
+)
+
+type NetConnSocket interface {
+	net.Conn
+
+	ReadWithAttributes(p []byte, attr *PacketAttributes) (n int, err error)
```
JoeTurki: I think it would be better if we unify attribute type in both  `ReadWithAttributes` functions.

packetio/buffer_test.go
```diff
@@ -257,63 +259,63 @@ func TestBufferLimitSize(t *testing.T) {
...
+	buffer.SetLimitSize(17)
 
 	assert.Equal(0, buffer.Size())
 
 	// Write twice
 	n, err := buffer.Write([]byte{0, 1})
 	assert.NoError(err)
 	assert.Equal(2, n)
-	assert.Equal(4, buffer.Size())
+	assert.Equal(6, buffer.Size())
```
JoeTurki: This is related to the comment about buffer.Write. we shouldn't change the test.

packetio/buffer.go
```diff
@@ -154,41 +216,61 @@ func (b *Buffer) Write(packet []byte) (int, error) { //nolint:cyclop
...
+
+	// increment the number of packets in buffer
 	b.count++
 
+	// store the length of the attributes segment
+	b.writeLengthHeaders(aLen)
+
+	if attr != nil {
+		// store the attributes buffer itself
+		b.writeFromInputBuffer(attr.Marshal())
```
JoeTurki: Can we make it so we call attr.Marshal once?

JoTurk (commented):
This is exciting, thank you! And sorry for all the comments, I tried to avoid anything super nitpicky. adding new APIs is always a bit hard .
It'd be great to also have tests covering the attributes path.

packetattributes.go
```diff
@@ -0,0 +1,57 @@
...
+const attrMaxLen = 1
+
+type ECN uint8
+
+type PacketAttributesBuffer interface {
+	// for serializing
+	Marshal() []byte
+
+	// for de-serializing. The bytes will be copied into the returned buffer
+	GetBuffer() []byte
```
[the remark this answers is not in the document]
penhauer: Yes yes I'm familiar with that. My idea was, GetBuffer returns the attributes buffer so bytes can be directly copied into the PacketAttributes' buffer instead of copying first into a temp buffer and then calling UnMarshall. The reason is, through a packet's journey, a packet is written/read to the packet buffer multiple times.

If you think there's a better option or UnMarshall is a good enough name, I'm fine with that.

connsocket.go
penhauer: I think having a generic packet attributes type with variable buffer length would address this comment right?

packetio/buffer_test.go
penhauer: I explained that buffer's write writes two segments in the buffer per packet. So this has to change. Also if you're not confident that it works, I have added a test at the end where I write some packets **without** attributes and read the same packets in order **with** attributes. It doesn't break any functionality.

packetio/buffer.go
```diff
@@ -220,17 +292,18 @@ func (b *Buffer) Read(packet []byte) (n int, err error) { //nolint:gocognit,cycl
...
+			b.advanceHead(count)
+
+			// read the attributes segment
+			aLen := b.getSegmentLength()
+			if aLen > 0 {
+				if attr != nil {
+					b.writeToInputBuffer(attr.GetBuffer(), aLen)
+				}
+				b.advanceHead(aLen)
 			}
```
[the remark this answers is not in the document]
penhauer: Yes you're right. I missed this one.

packetio/buffer.go
penhauer: Yes

packetio/buffer.go
JoeTurki: We're not the only users of pion/transport, even if receive and send works for us. If a user writes 2 bytes using Write and now it writes 4 bytes. this is a breaking behavior, even if the API itself didn't change.
We can add like a private flag or something `hasAttributes`.

connsocket.go
JoeTurki: yeah a generic packet attributes will be cool, good idea.

packetio/buffer.go
JoeTurki: This is not correct, because in packetio.Buffer the internal byte count directly affects observable behavior through:

1. Size()
2. SetLimitSize()
3. SetLimitCount()
4. backpressure via Write() returning ErrFull when the buffer is full, now we have extra bytes.
5. how grow() decides the allocation size
6. how many packets fit before being dropped/rejected

So even if Write() returns the same n and Read() returns the same n, the extra internal bytes materially change the buffer's behavior, it's not just about size.

connsocket.go
```diff
@@ -0,0 +1,95 @@
...
+
+type PacketConnToPacketConnSocket struct {
+	conn net.PacketConn
+}
+
+func NewPacketConnToPacketConnSocket(conn net.PacketConn) *PacketConnToPacketConnSocket {
+	return &PacketConnToPacketConnSocket{
+		conn: conn,
+	}
+}
```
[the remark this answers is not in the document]
JoeTurki: Can't this just be a helper-cast function that simply returns an interface, err without exposing a new public struct?

JoTurk:
https://github.com/pion/transport/pull/399

Issue #27, opened 2019-07-17
vnet: Support TCP


enobufs:
Current pion only supports UDP, but we will likely to support TCP as well. To help our development, we'd like to see vnet also support TCP.

See also: pion/webrtc#712

enobufs:
@Sean-Der can we add this to the project 2.2.0? I can help.

dgunay:
It looks like this blocks https://github.com/pion/ice/pull/394/

enobufs:
TODO: look into https://github.com/xtaci/kcp-go

stv0g:
TODO: look into https://github.com/google/gvisor/tree/master/pkg/tcpip

I think we could use the gVisor userspace networking stack to also replace the UDP transport, not only TCP.
But I am not sure to which extend the gVisor netstack also implements routing and NATting..

PR #375, opened 2026-01-14
Added TCP support to vnet

sirzooro:
Added TCP support to vnet. Github Copilot/GPT-5.2 did a good job :)

Ref: https://github.com/pion/transport/issues/27

### DIFF
--- a/vnet/tcp_conn.go
+++ b/vnet/tcp_conn.go
@@ -4,44 +4,502 @@
 package vnet
 
 import (
+	"errors"
 	"io"
 	"net"
+	"sync"
 	"time"
 
+	"github.com/pion/logging"
 	"github.com/pion/transport/v4"
+)
+
+type tcpState uint8
+
+const (
+	tcpStateInit tcpState = iota
+	tcpStateSynSent
+	tcpStateSynReceived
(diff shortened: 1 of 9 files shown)


vnet/tcp_conn.go
```diff
@@ -0,0 +1,400 @@
...
+	c.readTimer.Reset(d)
+
+	return nil
+}
+
+// SetWriteDeadline sets the deadline for future Write calls.
+func (c *TCPConn) SetWriteDeadline(t time.Time) error {
+	// Writes do not block in vnet.
+	return nil
+}
```
JoTurk, line 497: We have had many bugs due to deadlines with TCP and i think tcp vnet should be able to emulate real writes, as well as blocked writes. Non-blocking writes kinda defeat the purpose of testing tcp with vnet for many cases.

(I'm just going through this fast before i eventually review this in few days lol)

sirzooro: Done, added deadlines, ACKs and small internal queue of 10 packets.

sirzooro:
I am going to split this big PR into few smaller ones, to make review easier.

PR #377, opened 2026-02-04
Resolve the empty address to INADDR_ANY in VNet

rg0now:
#### Description

So far calling `VNet.ListenPacket("udp4", ":1234")` with the shorthand empty address has failed by "host not found" in `VNet.ResolveIPAddr`. 

This PR makes sure the empty address resolves to `INADDR_ANY`.

### DIFF
--- a/vnet/net.go
+++ b/vnet/net.go
@@ -320,6 +320,10 @@ func (v *Net) Dial(network string, address string) (net.Conn, error) {
 func (v *Net) ResolveIPAddr(_, address string) (*net.IPAddr, error) {
 	var err error
 
+	if address == "" {
+		address = "0.0.0.0"
+	}
+
 	// Check if host is a domain name
 	ip := net.ParseIP(address)
 	if ip == nil { //nolint:nestif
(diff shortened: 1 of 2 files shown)


vnet/net.go
```diff
@@ -320,6 +320,10 @@ func (v *Net) Dial(network string, address string) (net.Conn, error) {
 func (v *Net) ResolveIPAddr(_, address string) (*net.IPAddr, error) {
 	var err error
 
+	if address == "" {
+		address = "0.0.0.0"
```
JoTurk, line 324: I think it will be cleaner if we use `net.IPv4zero` also we should prob handle ipv6.

JoTurk (approved):
thank you.

rg0now:
@JoTurk Sorry for this taking so long. Redid the patch, also fixed `ResolveUDPAddr` and `ResolveTCPAddr` along the way. All tests pass on my side...

JoTurk:
@rg0now no worries :)

(I think the CI is failing because github had issues, Let me try to restart it).

JoTurk (approved):
I think we should follow up by updating dial and listen helpers to accept udp6. I'll merge this and open a PR.

PR #383, opened 2026-05-28
Support udp6 dial

JoTurk:
#### Description
Support UDP6 dial and listen, a follow-up on https://github.com/pion/transport/pull/377.

### DIFF
--- a/vnet/net.go
+++ b/vnet/net.go
@@ -143,7 +143,11 @@ func (v *Net) getAllIPAddrs(ipv6 bool) []net.IP {
 				continue
 			}
 
-			if !ipv6 {
+			if ipv6 {
+				if ip.To4() == nil {
+					ips = append(ips, ip)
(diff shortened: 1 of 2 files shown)


rg0now (approved):

vnet/net.go
```diff
@@ -143,7 +143,11 @@ func (v *Net) getAllIPAddrs(ipv6 bool) []net.IP {
 				continue
 			}
 
-			if !ipv6 {
+			if ipv6 {
+				if ip.To4() == nil {
```
adrianosela, line 147: Just a heads up that `To4()` returns non-nil for IPv4-mapped IPv6s e.g. `::ffff:192.168.1.1`.
Not sure if we are handling that correctly, looks like we just dont populate `ips` with those addresses.

adrianosela:
LGTM!

PR #387, opened 2026-07-05
Fix TCP chunk: seqNum/ackNum and Clone

sirzooro:
#### Description

Self-contained improvements to the existing `chunkTCP` struct.
- Add `seqNum` and `ackNum` fields (replacing commented-out stubs).
- Fix `Clone()` to copy `flags`, `seqNum`, `ackNum`, `tag`, and `duplicate` — previously these were silently dropped on clone.
- Use the `tcp` package constant in `Network()` instead of raw string `"tcp"`.

Split out from https://github.com/pion/transport/pull/375

### DIFF
--- a/vnet/chunk.go
+++ b/vnet/chunk.go
@@ -81,7 +81,7 @@ type chunkIP struct {
 	duplicate     bool
 }
 
-func (c *chunkIP) setTimestamp() time.Time {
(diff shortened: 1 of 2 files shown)


JoTurk (approved):

sirzooro:
Thanks

PR #392, opened 2026-08-03
New allocation-free replay detector API

kcaffrey:
The current `replaydetector.ReplayDetector` API allocates for each inbound packet in order to create and return the closure to accept the packet. In order to avoid this allocation, a new two-phase API is introduced that has separate check and accept methods.

The detectors returned by `New` and `WithWrap` implement both the new and legacy APIs. The behavior of the legacy API is unchanged. The new API differs for the wrapped variant with regard to out-of-order accepts; the legacy API could end up erroneously accepting a replayed packet in this case while the new API will not.

The new API uses a "token" to pass to accept (as opposed to, e.g., having Accept take a sequence number) in order to allow the same no-op on rejected packets pattern that the legacy API supported. Factory methods for creating accepted and rejected tokens exist to allow users to create custom replay detectors while making it difficult (or at least obvious) to misuse the API.

### DIFF
--- a/replaydetector/replaydetector.go
+++ b/replaydetector/replaydetector.go
@@ -13,6 +13,52 @@ type ReplayDetector interface {
 	Check(seq uint64) (accept func() bool, ok bool)
 }
 
+// Token records the outcome of a replay check for one sequence number.
+// The zero value is inert: accepting it does nothing.
+type Token struct {
+	seq    uint64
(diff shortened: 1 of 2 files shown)


Sean-Der:
Your a wizard @kcaffrey :)

Sean-Der (approved):

PR #395, opened 2026-08-30
Switch goroutines for contexts in packetconn

noboruma:
Remove goroutines in Write/Read hot paths from packetconn


netctx/packetconn.go
```diff
@@ -58,47 +60,34 @@ func NewPacketConn(pconn net.PacketConn) PacketConn {
...
+	if p.ctx.Err() != nil {
+		return 0, nil, net.ErrClosed
 	}
-	if err2, ok := errSetDeadline.Load().(error); ok && err == nil && err2 != nil {
-		err = err2
+	if ctx.Err() != nil {
+		_ = p.nextConn.SetReadDeadline(time.Time{})
+
+		return 0, nil, ctx.Err()
 	}
```
JoTurk: Shouldn't remove the `n == 0` check from the old implementation, this  breaks the net.PacketConn contract if packets arrive at the same time of the cancellation.

> Callers should always process
	// the n > 0 bytes returned before considering the error err.
	// ReadFrom can be made to time out and return an error after a
	// fixed time limit; see SetDeadline and SetReadDeadline.

https://pkg.go.dev/net#PacketConn

netctx/packetconn.go
```diff
@@ -58,47 +60,34 @@ func NewPacketConn(pconn net.PacketConn) PacketConn {
...
-		case <-done:
-		}
-	}()
+	combinedCtx, cancel := context.WithCancel(ctx)
+	defer cancel()
+
+	detachDeadline := context.AfterFunc(combinedCtx, func() {
+		_ = p.nextConn.SetReadDeadline(veryOld)
+	})
+	defer detachDeadline()
```
JoTurk: This can race and cancel next read I think we should make sure that setDeadline calls are not defered if we keep the defer / AfterFunc we should make sure that AfterFunc calls are synced.

netctx/packetconn.go
```diff
@@ -58,47 +60,34 @@ func NewPacketConn(pconn net.PacketConn) PacketConn {
...
-				errSetDeadline.Store(err)
-			}
-		case <-done:
-		}
-	}()
+	combinedCtx, cancel := context.WithCancel(ctx)
+	defer cancel()
+
+	detachDeadline := context.AfterFunc(combinedCtx, func() {
+		_ = p.nextConn.SetReadDeadline(veryOld)
```
JoTurk: nit: maybe we shouldn't ignore deadline errors?

JoTurk (commented):
Looks great thank you, I also tested it locally :)

Just 2 comments and one nit.

netctx/packetconn.go
noboruma: So actually I gave that one some thoughts because it was specifically handled in the previous impl. My conclusion here is that failing at the `SetReadDeadline` stage should only happen if the connection is closed already. So we don't really care at that stage. Now this holds true for standard implementations, in the case the connection was not closed and the `SetReadDeadline` truly failed (which in theory should never happen unless the underlying implementation is buggy) we would be in much more trouble since that operation could hang forever (which was also the case in the previous impl since we only returned the `err` after the read finished).

I think a better alternative here would be to simply `panic` in case of error so it gives a chance to recover?

netctx/packetconn.go
noboruma: Good catch!
We need to cancel the association earlier, will rework that

netctx/packetconn.go
noboruma: My reading here is that we can return both `n` and the `err` at the same time and it's caller responsibility to handle both.
Returning `n` should do the trick

netctx/packetconn.go
noboruma: Actually realized we cannot do that properly because `AfterFunc` would panic on a different goroutine - reworking the whole `SetReadDeadline` instead

noboruma: Here I am opting for a closure of the whole connection so that we avoid hanging

JoTurk:
@noboruma hello, can you address the copilot review? i also suspect few parts but I need to test to confirm. I'll tag v5 now and we can get this for v5.0.1 since it doesn't have breaking changes :)

PR #396, opened 2026-09-02
Add AfterTimeout to Deadline

noboruma:
Add `AfterFunc` to `deadline` to allow for continuation on when deadline are reached.
Convenient for avoiding long running `goroutine`s with connection contexts.

This change is needed for: https://github.com/pion/dtls/pull/1096

### DIFF
--- a/deadline/deadline.go
+++ b/deadline/deadline.go
@@ -30,6 +30,8 @@ type Deadline struct {
 	deadline time.Time
 	state    deadlineState
 	pending  uint8
+	cbs      map[int]func()
+	nextCbID int
 }
 
 // New creates new deadline timer.
@@ -47,11 +49,41 @@ func (d *Deadline) timeout() {
 		return
 	}
 
-	d.state = deadlineExceeded
-	done := d.done
+	d.fire()
 	d.mu.Unlock()
+}
 
-	close(done)
+// AfterTimeout attaches a callback to the deadline.
+// The added callback will be triggered when the deadline is met.
+// The callback can be detached via the returned function.
+// This function mimics context.AfterFunc behavior.
+func (d *Deadline) AfterTimeout(cb func()) func() bool {
+	d.mu.Lock()
+	defer d.mu.Unlock()
+	if d.state == deadlineExceeded {
+		go cb()
+
+		return func() bool { return false }
+	}
+	d.nextCbID++
+	usedID := d.nextCbID
+	if d.cbs == nil {
+		d.cbs = map[int]func(){}
+	}
+	d.cbs[d.nextCbID] = cb
+	cancel := func() bool {
+		d.mu.Lock()
+		defer d.mu.Unlock()
+		if _, has := d.cbs[usedID]; has {
+			delete(d.cbs, usedID)
+
+			return true
+		}
+
+		return false
+	}
+
+	return cancel
 }
 
 // Set new deadline. Zero value means no deadline.
@@ -89,8 +121,16 @@ func (d *Deadline) Set(setTo time.Time) {
 	}
 
 	d.pending--
+	d.fire()
+}
+
+func (d *Deadline) fire() {
 	d.state = deadlineExceeded
 	close(d.done)
+	for _, cb := range d.cbs {
+		go cb()
+	}
+	clear(d.cbs)
 }
 
 // Done receives deadline signal.
--- a/deadline/deadline_test.go
+++ b/deadline/deadline_test.go
@@ -5,6 +5,7 @@ package deadline
 
 import (
 	"context"
+	"sync/atomic"
 	"testing"
 	"time"
 
@@ -92,6 +93,88 @@ func TestDeadline(t *testing.T) {
 		calls := collectCh(ch, 2, 60*time.Millisecond)
 		expectedCalls := []byte{0}
 		assert.Equal(t, expectedCalls, calls, "Wrong order of deadline signal")
+	})
+
+	t.Run("DeadlineAfterTimeoutExceedFuture", func(t *testing.T) {
+		now := time.Now()
+
+		d := New()
+		var trigger atomic.Int32
+		d.AfterTimeout(func() {
+			trigger.Add(1)
+		})
+		d.Set(now.Add(10 * time.Millisecond))
+		<-time.After(20 * time.Millisecond)
+		d.Set(now.Add(10 * time.Millisecond))
+		<-time.After(20 * time.Millisecond)
+
+		assert.Equal(t, int32(1), trigger.Load(), "Function triggered wrong number of time")
+	})
+
+	t.Run("DeadlineAfterTimeoutExceedPast", func(t *testing.T) {
+		now := time.Now()
+
+		d := New()
+		var trigger atomic.Int32
+		d.AfterTimeout(func() {
+			trigger.Add(1)
+		})
+		d.Set(now.Add(10 * time.Millisecond))
+		d.Set(now.Add(-10 * time.Millisecond))
+		<-time.After(20 * time.Millisecond)
+
+		assert.Equal(t, int32(1), trigger.Load(), "Function triggered wrong number of time")
+	})
+
+	t.Run("DeadlineAfterTimeoutAlreadyExceeded", func(t *testing.T) {
+		now := time.Now()
+
+		d := New()
+		var trigger atomic.Int32
+		d.Set(now.Add(10 * time.Millisecond))
+		d.Set(now.Add(-10 * time.Millisecond))
+		<-time.After(20 * time.Millisecond)
+		d.AfterTimeout(func() {
+			trigger.Add(1)
+		})
+		<-time.After(20 * time.Millisecond)
+
+		assert.Equal(t, int32(1), trigger.Load(), "Function triggered wrong number of time")
+	})
+
+	t.Run("DeadlineAfterTimeoutDetach", func(t *testing.T) {
+		now := time.Now()
+
+		d := New()
+		var trigger atomic.Int32
+		detach := d.AfterTimeout(func() {
+			trigger.Add(1)
+		})
+		ret := detach()
+		d.Set(now.Add(10 * time.Millisecond))
+		<-time.After(20 * time.Millisecond)
+
+		assert.Equal(t, int32(0), trigger.Load(), "Function triggered wrong number of time")
+		assert.Equal(t, true, ret, "The detach function returned wrong value")
+	})
+
+	t.Run("DeadlineAfterTimeoutMultiCallbacks", func(t *testing.T) {
+		now := time.Now()
+
+		d := New()
+		var trigger atomic.Int32
+		d.AfterTimeout(func() {
+			trigger.Add(1)
+		})
+		d.AfterTimeout(func() {
+			trigger.Add(1)
+		})
+		assert.Equal(t, int32(0), trigger.Load(), "Function triggered wrong number of time")
+
+		d.Set(now.Add(10 * time.Millisecond))
+		<-time.After(20 * time.Millisecond)
+
+		assert.Equal(t, int32(2), trigger.Load(), "Function triggered wrong number of time")
 	})
 }
 


deadline/deadline.go
```diff
@@ -73,6 +105,7 @@ func (d *Deadline) Set(setTo time.Time) {
 	if setTo.IsZero() {
 		d.pending--
 		d.state = deadlineStopped
+		clear(d.cbs)
```
paulwe: **Blocker.** This makes `Set(time.Time{})` permanently discard every registered callback.

`SetReadDeadline(time.Time{})` is the documented way to disable a deadline (`net.go`, and `test/bridge.go:142` does it), so ordinary code silently unregisters all continuations. A later `Set(future)` that genuinely expires then fires nothing.

Worse, the caller can't detect it: the `cancel` func returned by `AfterFunc` now returns `false`, which is indistinguishable from "already fired." And via the `afterFuncer` path, any `context.WithCancel(d)` child created before this call never cancels.

If the intent is "no deadline means the callbacks can never fire," that's incompatible with being a `Context` — the child has no way to learn its parent will never cancel. I'd keep the callbacks registered across `Set(zero)`.

deadline/deadline.go
```diff
@@ -49,11 +52,40 @@ func (d *Deadline) timeout() {
...
 }
 
+// AfterFunc attaches a function to the deadline.
+// The functions will be triggered on deadline exceeded.
+// If the deadline is reset, the functions are skipped.
+// Attached functions are only triggered once.
+func (d *Deadline) AfterFunc(cb func()) func() bool {
+	d.mu.Lock()
+	defer d.mu.Unlock()
+	d.cbs[d.nextCbID] = cb
```
paulwe: **Blocker.** If the deadline is already exceeded, this stores `cb` and returns without ever invoking it. `context.AfterFunc` fires immediately in that case.

It's also timing-dependent rather than consistent: if the timer has fired but `timeout()` is still blocked on `d.mu`, a callback registered at that moment *does* run. Same logical state, two different outcomes.

This is precisely the window `context.propagateCancel` operates in — it does `select { case <-done: ...; default: }` and only then calls `a.AfterFunc(...)`. If the deadline expires between those, the child registers into a map that was just cleared and never cancels.

```go
if d.state == deadlineExceeded {
	go cb()
	return func() bool { return false }
}
```

Separately: a `nil` cb is accepted here and panics later in a detached goroutine, killing the process with a stack that doesn't name this call site. I know `context.AfterFunc` doesn't check either — but there a nil always surfaces, since every path reaches `go a.f()`. Here, combined with the bug above, `d.AfterFunc(nil)` on an expired deadline produces no symptom at all. Same input, silent or fatal depending on timing.

deadline/deadline.go
```diff
@@ -49,11 +52,40 @@ func (d *Deadline) timeout() {
 
 	d.state = deadlineExceeded
 	done := d.done
+	for _, cb := range d.cbs {
+		go cb()
+	}
+	clear(d.cbs)
```
paulwe: Two things here.

**Ordering differs from the other expiry path.** Here the callbacks are spawned, then `d.mu.Unlock()` (L59), then `close(done)` (L61). In `Set` the close happens *before* the dispatch (L126 vs L127). So on this path a callback can run while `d.Done()` is still open — but `Err()` already returns `DeadlineExceeded` the moment L59 releases the lock. A context child that cancels off `parent.Err()` is then observably canceled before its parent. Dispatching after `close(done)` in both paths removes it.

**Spawning under the lock.** N `go` statements while holding `d.mu` delays every concurrent `Set`/`Done`/`Err`, and lets a callback that starts on another P block on the lock its own dispatcher still holds. Snapshot into a slice under the lock, spawn after `Unlock`. (Per-callback goroutines are right and match `context.AfterFunc` — no argument there, and the dispatch order genuinely doesn't matter since they race regardless.)

deadline/deadline.go
```diff
@@ -49,11 +52,40 @@ func (d *Deadline) timeout() {
...
 	d.mu.Unlock()
 
 	close(done)
 }
 
+// AfterFunc attaches a function to the deadline.
+// The functions will be triggered on deadline exceeded.
+// If the deadline is reset, the functions are skipped.
```
paulwe: "reset" has no referent in the API — there's no `Reset` method, only `Set`. I initially read it as `Timer.Reset` semantics (re-arm to a new time), under which it's backwards: `Set(future)` *keeps* the callbacks and fires them at the new deadline. Read as `Set(time.Time{})` it's accurate.

That ambiguity is the problem — the two `Set` paths do opposite things to the registry, so it's worth naming them:

```go
// AfterFunc registers cb to run on its own goroutine when the deadline is
// exceeded. Setting a new future deadline re-arms cb; it is not skipped.
// The returned func reports whether it removed cb before it ran.
```

Also worth stating that `cb` runs on its own goroutine, and what the `bool` means.

deadline/deadline.go
```diff
@@ -49,11 +52,40 @@ func (d *Deadline) timeout() {
...
+// Attached functions are only triggered once.
+func (d *Deadline) AfterFunc(cb func()) func() bool {
```
paulwe: Worth a line noting this signature deliberately satisfies `context`'s (unexported) `afterFuncer`, so derived contexts take the `propagateCancel` fast path instead of the goroutine fallback.

The coupling is compile-silent: drop the `bool`, add an `error`, and the type quietly falls back to a goroutine per child with no build error and no failing test. A comment plus the integration test suggested below pins it.

deadline/deadline.go
```diff
@@ -30,12 +30,15 @@ type Deadline struct {
...
 	pending  uint8
+	cbs      map[int]func()
+	nextCbID int
 }
 
 // New creates new deadline timer.
 func New() *Deadline {
 	return &Deadline{
 		done: make(chan struct{}),
+		cbs:  map[int]func(){},
```
paulwe: Every `Deadline` now pays a map allocation at construction for a feature almost none of them use — `test/bridge.go:51-52` and dtls allocate two per connection.

Allocating lazily in `AfterFunc` also fixes `var d deadline.Deadline; d.AfterFunc(f)`, which currently panics with "assignment to entry in nil map":

```go
if d.cbs == nil {
	d.cbs = map[int]func(){}
}
```

deadline/deadline.go
```diff
@@ -49,11 +52,40 @@ func (d *Deadline) timeout() {
...
 
+// AfterFunc attaches a function to the deadline.
+// The functions will be triggered on deadline exceeded.
+// If the deadline is reset, the functions are skipped.
+// Attached functions are only triggered once.
+func (d *Deadline) AfterFunc(cb func()) func() bool {
+	d.mu.Lock()
+	defer d.mu.Unlock()
+	d.cbs[d.nextCbID] = cb
+	usedID := d.nextCbID
```
paulwe: `d.nextCbID` is read at L71, again here, and incremented at L84 — straddling the closure literal. Moving the increment up next to its siblings, which looks like a safe tidy-up, makes `cancel` delete a key that was never inserted: it silently stops working and every callback fires.

```go
id := d.nextCbID
d.nextCbID++
d.cbs[id] = cb
```

deadline/deadline.go
```diff
@@ -91,6 +124,10 @@ func (d *Deadline) Set(setTo time.Time) {
 	d.pending--
 	d.state = deadlineExceeded
 	close(d.done)
+	for _, cb := range d.cbs {
+		go cb()
+	}
+	clear(d.cbs)
```
paulwe: Same four lines as L55-58, but positioned differently relative to `close(done)` — which is exactly the inconsistency noted there. Any future change to expiry has to land in both places and will drift again.

Extracting one `func (d *Deadline) expire(done chan struct{})` that closes `done` and drains `d.cbs`, called from both sites, makes the paths provably identical.

deadline/deadline_test.go
```diff
@@ -93,6 +94,68 @@ func TestDeadline(t *testing.T) {
...
+		now := time.Now()
+
+		d := New()
+		var trigger atomic.Int32
+		d.AfterFunc(func() {
+			trigger.Add(1)
+		})
+		d.Set(now.Add(10 * time.Millisecond))
+		<-time.After(20 * time.Millisecond)
+		d.Set(now.Add(10 * time.Millisecond))
```
paulwe: `now` is captured at L99 and L107 sleeps 20ms, so this is ~10ms in the *past* by the time it runs — it takes `Set`'s immediate-exceed branch, not the timer. The scenario the name promises (expire via timer, re-arm, expire via timer again) isn't covered; use `time.Now().Add(...)` here.

More generally the 10ms/20ms margins across all four subtests are tight for a loaded CI runner.

deadline/deadline_test.go
```diff
@@ -93,6 +94,68 @@ func TestDeadline(t *testing.T) {
...
+
+	t.Run("DeadlineAfterFuncDetach", func(t *testing.T) {
+		now := time.Now()
+
+		d := New()
+		var trigger atomic.Int32
+		detach := d.AfterFunc(func() {
+			trigger.Add(1)
+		})
+		detach()
```
paulwe: The `bool` is discarded here and never asserted anywhere in the suite, so the return contract is untested — `cancel` could always return `false` and everything still passes.

Worth asserting `true` here, and `false` on a second call.

deadline/deadline_test.go
```diff
@@ -93,6 +94,68 @@ func TestDeadline(t *testing.T) {
...
+			trigger.Add(1)
+		})
+		d.Set(now.Add(10 * time.Millisecond))
+		d.Set(now.Add(-10 * time.Millisecond))
+		<-time.After(20 * time.Millisecond)
+
+		assert.Equal(t, int32(1), trigger.Load(), "Function triggered wrong number of time")
+	})
+
+	t.Run("DeadlineAfterFuncCancel", func(t *testing.T) {
```
paulwe: Misnamed — this tests `Set(time.Time{})`, not the returned cancel func (that's `DeadlineAfterFuncDetach` below). Maybe `DeadlineAfterFuncSetZero`.

It also locks in the `clear(d.cbs)` behavior flagged on `deadline.go:108` as intended, so it'll need to change alongside that fix.

deadline/deadline_test.go
```diff
@@ -93,6 +94,68 @@ func TestDeadline(t *testing.T) {
 		expectedCalls := []byte{0}
 		assert.Equal(t, expectedCalls, calls, "Wrong order of deadline signal")
 	})
+
+	t.Run("DeadlineAfterFuncExceedFuture", func(t *testing.T) {
```
paulwe: These four subtests are the same eight-line skeleton differing only in the `Set` calls and expected count — `dupl` already forces `//nolint` on the sibling tests in this file. A table of `{name, ops func(*Deadline, time.Time), want int32}` collapses them and makes the missing cases one line each.

The gaps that matter most, none currently covered:

- **Register while already exceeded** — catches the `deadline.go:71` blocker in three lines: `d.Set(now.Add(-time.Millisecond)); d.AfterFunc(f); <-time.After(20*time.Millisecond)`, expect 1.
- **`context.WithCancel(d)`'s child cancels when `d` expires** — the integration this PR exists to enable, and the only thing that pins the compile-silent `afterFuncer` coupling.
- **Concurrent `AfterFunc` while expiring**, under `-race`.
- **Multiple callbacks** — everything here registers exactly one.

deadline/deadline.go
paulwe, line 108: clearing `d.cbs` here breaks the `context.afterFuncer` contract - there's no way to disambiguate "already fired" from "discarded" when the cancel func returns false. it also orphans derived contexts eg. from `context.WithCancel(d)`.

deadline/deadline.go
paulwe, line 71: when the deadline is already exceeded cb should fire immediately

if the local var is used in every read reordering later can't break things...
```go
d.nextCbID++
usedID := d.nextCbID
d.cbs[usedID] = cb
// ...
```

deadline/deadline.go
paulwe: Two things here.

**Ordering differs from the other expiry path.** Here the callbacks are spawned, then `d.mu.Unlock()` (L59), then `close(done)` (L61). In `Set` the close happens *before* the dispatch (L126 vs L127). So on this path a callback can run while `d.Done()` is still open — but `Err()` already returns `DeadlineExceeded` the moment L59 releases the lock. A context child that cancels off `parent.Err()` is then observably canceled before its parent. Dispatching after `close(done)` in both paths removes it.

**Spawning under the lock.** N `go` statements while holding `d.mu` delays every concurrent `Set`/`Done`/`Err`, and lets a callback that starts on another P block on the lock its own dispatcher still holds. Snapshot into a slice under the lock, spawn after `Unlock`. (Per-callback goroutines are right and match `context.AfterFunc` — no argument there, and the dispatch order genuinely doesn't matter since they race regardless.)

deadline/deadline.go
paulwe: "reset" has no referent in the API — there's no `Reset` method, only `Set`. I initially read it as `Timer.Reset` semantics (re-arm to a new time), under which it's backwards: `Set(future)` *keeps* the callbacks and fires them at the new deadline. Read as `Set(time.Time{})` it's accurate.

That ambiguity is the problem — the two `Set` paths do opposite things to the registry, so it's worth naming them:

```go
// AfterFunc registers cb to run on its own goroutine when the deadline is
// exceeded. Setting a new future deadline re-arms cb; it is not skipped.
// The returned func reports whether it removed cb before it ran.
```

Also worth stating that `cb` runs on its own goroutine, and what the `bool` means.

deadline/deadline.go
paulwe: the comment string should probably explain that this implements the `context.afterFuncer` interface so someone doesn't change the signature later and silently break it

deadline/deadline.go
paulwe, line 41: unless most `Deadline` have callbacks registered going forward this should be initialized lazily in `AfterFunc`

deadline/deadline.go
paulwe: Same four lines as L55-58, but positioned differently relative to `close(done)` — which is exactly the inconsistency noted there. Any future change to expiry has to land in both places and will drift again.

Extracting one `func (d *Deadline) expire(done chan struct{})` that closes `done` and drains `d.cbs`, called from both sites, makes the paths provably identical.

deadline/deadline_test.go
paulwe: `now` is captured at L99 and L107 sleeps 20ms, so this is ~10ms in the *past* by the time it runs — it takes `Set`'s immediate-exceed branch, not the timer. The scenario the name promises (expire via timer, re-arm, expire via timer again) isn't covered; use `time.Now().Add(...)` here.

More generally the 10ms/20ms margins across all four subtests are tight for a loaded CI runner.

deadline/deadline_test.go
paulwe, line 153: no test currently checks that the `bool` return value is correct

deadline/deadline_test.go
paulwe, line 129: the name doesn't match the comment above. also this pins the questionable `Set(time.Time{})` behavior...

deadline/deadline_test.go
paulwe, line 98: these can probably be a table test eg.`{name, ops func(*Deadline, time.Time), want int32}` with additional cases to cover `AfterFunc(f)` while already exceeded, `context.WithCancel(d)` cancel propagation, and multiple callbacks.

deadline/deadline.go
```diff
@@ -49,11 +52,40 @@ func (d *Deadline) timeout() {
...
 	d.mu.Unlock()
 
 	close(done)
 }
 
+// AfterFunc attaches a function to the deadline.
+// The functions will be triggered on deadline exceeded.
+// If the deadline is reset, the functions are skipped.
+// Attached functions are only triggered once.
```
paulwe, line 67: the wording is unclear because there is no "reset" function and `Set(zero)` and `Set(futureTime)` have opposite effects on the pending callbacks. the cancel func semantics should be documented. it should be stated that this implements the `context.afterFuncer` interface so someone doesn't change the signature later and silently break it

deadline/deadline.go
```diff
@@ -49,11 +52,40 @@ func (d *Deadline) timeout() {
 
 	d.state = deadlineExceeded
 	done := d.done
+	for _, cb := range d.cbs {
+		go cb()
+	}
+	clear(d.cbs)
 	d.mu.Unlock()
 
 	close(done)
```
paulwe, line 61: my code here is wrong: [context.go:111](https://github.com/golang/go/blob/master/src/context/context.go#L111)
```go
// If Done is not yet closed, Err returns nil.
```
`close(d.done)` should move into the critical section above the dispatch loop so callbacks cannot observe `done` unclosed.

deadline/deadline.go
paulwe, line 130: this and `timeout` should share a helper to prevent the order from drifting

deadline/deadline_test.go
```diff
@@ -93,6 +94,68 @@ func TestDeadline(t *testing.T) {
...
+
+		d := New()
+		var trigger atomic.Int32
+		d.AfterFunc(func() {
+			trigger.Add(1)
+		})
+		d.Set(now.Add(10 * time.Millisecond))
+		<-time.After(20 * time.Millisecond)
+		d.Set(now.Add(10 * time.Millisecond))
+		<-time.After(20 * time.Millisecond)
```
paulwe, line 109: 10/20ms margins will probably flake in CI. also double check that this is taking the expected path...

JoTurk:
@noboruma I thought about this a lot and the dtls pr l, I think the cleanest approach is to avoid implementing context.afterFuncer on Deadline. Could this instead expose something like OnDeadline or AfterDeadline have DTLS use that directly? This also give us the freedom to use a different notify design other than a callback.
I'll also push to your other dtls pr to fix some minor issues and try to get it merged today and I can help with this.

deadline/deadline.go
noboruma: Fair enough, if we consider putting to 0 was akin to be indefinitely running, however this indeed breaks the case where the deadline would be reset after that.

However I was under the impression here we are supporting the `stopped` state since we have this `deadlineStopped` state. Maybe we should not support stopping then? (either we are `starting` or `started` or `exceeded`)

deadline/deadline.go
noboruma: Interesting, I now wonder if this is the root cause of the issue that triggered this whole PR.
I was using this struct directly with `context.AfterFunc` but was hitting panics because of `Err()` returning `nil`.

Let me double check that once again, maybe we can discard the whole PR and only keep the fix

noboruma: So actually by design `Deadline` cannot be used for `context.AfterFunc` because of it's mutable nature, whereas `context.AfterFunc` relies on the fact the context cannot change once it's "done".
So we still need our own API, as proposed

deadline/deadline_test.go
noboruma: I have added tests for already exceeded and multi callbacks
Cancel is out of `Deadline` scope if I am not mistaken?
That's the whole reason we need `AfterFunc`-like behavior actually

Happy to move the tests into a table, but does it not make error tracing harder?

deadline/deadline.go
noboruma: I have reworded the whole accordingly (and also renamed the function as per Jo's comment)

deadline/deadline_test.go
noboruma: Ideally we should be able to control a fake clock but I have simply copied from above tests.
Maybe I can do better with channel synchronization instead, otherwise do you suggest higher timers?

deadline/deadline.go
paulwe: it's worse than i originally thought - implementations of `context.Context` must be monotonic state machines. stdlib can panic when a context is derived from a `Deadline` that is subsequently reset whether it implements `context.afterFuncer` or not. here is a failing test demonstrating the bug: [gist](https://gist.github.com/paulwe/345e2466b9a00ae6165474365f6ca42e)

deadline/deadline_test.go
paulwe: usually 100-200ms is enough slack for oversubscribed gh test runners. if the numbers you chose are found elsewhere in the package i retract my suggestion though

JoTurk: We should start using synctest for time tests like this, it should be beind a 1.25 go build flag.

deadline/deadline.go
JoTurk: @paulwe this is pre-existing, Set already violates the context.Context contract, and the panic can already happen, the linked dtls pr shouldn't trigger this panic with AfterTimeout but since this is public I support fixing this and pushing a new transport major release if needed.

paulwe: because the incompatibility is irreconcilable i don't think `Deadline` should pretend to implement `context.Context`. https://github.com/pion/transport/pull/397 implements a helper for deriving a real context from the deadline and https://github.com/pion/transport/pull/398 breaks the interface.

noboruma:
So with #397 this PR is now obsolete.
We can simply start using `context.AfterFunc` using the returned `Context()`.

PR #397, opened 2026-09-09
Add Deadline.Context for derived contexts

paulwe:
`deadline.Deadline` is revivable via `Set`, so it is not monotonic, and `context.Context` requires that it be:

> If Done is not yet closed, Err returns nil.
> If Done is closed, Err returns a non-nil error explaining why.
> After Err returns a non-nil error, successive calls to Err return the same error.

After a `Deadline` expires and is set again, `Err()` goes from `DeadlineExceeded` back to `nil` and `Done()` hands out a fresh open channel while earlier holders still see the old one closed — two observers of the same "context" disagreeing permanently.

That makes deriving a context from a `Deadline` unsafe. Both propagation paths in `context` read `parent.Err()` *after* observing `Done`:

```go
case <-parent.Done():
    child.cancel(false, parent.Err(), Cause(parent))
```

and `cancelCtx.cancel` panics on a nil error, before its already-cancelled early return. A `Set` landing in that window panics inside the standard library:

```
panic: context: internal error: missing cancel error
  context.(*cancelCtx).cancel(...)              context/context.go:552
  context.(*cancelCtx).propagateCancel.func2()  context/context.go:526
```

This reproduces on `main` today with no callback API in the tree.

### This PR

`Context()` returns a real cancel context scoped to the current deadline generation, so what callers hold is monotonic even though `Deadline` is not:

- memoized while the deadline is live — extending it keeps the same context
- cancelled when the deadline fires, with `context.DeadlineExceeded` as its `context.Cause`
- replaced on the next call after expiry, so a retired context stays cancelled forever

Because the returned value is a `*cancelCtx`, `parentCancelCtx` finds it and children register in its map — deriving 100 contexts spawns **0** goroutines, versus one watchdog goroutine per child for any parent `context` cannot recognise.

`timeout` and `Set` now share a `fire` helper, which also closes `done` inside the critical section and removes the pre-existing window where `Err()` reported `DeadlineExceeded` while `Done()` was still open.

Purely additive — `Set`, `Done`, `Err` and `Deadline` are untouched and the `context.Context` implementation is left in place. #398 removes that separately.

### Testing

`TestDeadlineContext` covers memoization, extend, stop, timer and past-deadline expiry, generation replacement with the retired context staying cancelled, derived-context cancellation, the zero-goroutine property, and concurrent `Context`/`Set`. Green under `-race`.

### DIFF
--- a/deadline/deadline.go
+++ b/deadline/deadline.go
@@ -14,10 +14,19 @@ import (
 type deadlineState uint8
 
 const (
-	deadlineStopped deadlineState = iota
+	deadlineSuspended deadlineState = iota
 	deadlineStarted
 	deadlineExceeded
 )
+
+//nolint:gochecknoglobals // A canceled context is immutable, so it is safe to share.
+var exceededContext context.Context
+
+func init() {
+	ctx, cancel := context.WithCancelCause(context.Background())
+	cancel(context.DeadlineExceeded)
+	exceededContext = ctx
+}
 
 var _ context.Context = (*Deadline)(nil)
 
@@ -75,6 +84,10 @@ func (d *Deadline) fire() context.CancelCauseFunc {
 func (d *Deadline) Context() context.Context {
 	d.mu.Lock()
 	defer d.mu.Unlock()
+
+	if d.state == deadlineExceeded {
+		return exceededContext
+	}
 	if d.ctx == nil {
 		d.ctx, d.cancel = context.WithCancelCause(context.Background())
 	}
@@ -106,7 +119,7 @@ func (d *Deadline) set(setTo time.Time) context.CancelCauseFunc {
 
 	if setTo.IsZero() {
 		d.pending--
-		d.state = deadlineStopped
+		d.state = deadlineSuspended
 
 		return nil
 	}
--- a/deadline/deadline.go
+++ b/deadline/deadline.go
@@ -24,9 +24,12 @@ var _ context.Context = (*Deadline)(nil)
 // Deadline signals updatable deadline timer.
 // Also, it implements context.Context.
 type Deadline struct {
-	mu       sync.RWMutex
-	timer    timer
-	done     chan struct{}
+	mu     sync.RWMutex
+	timer  timer
+	done   chan struct{}
+	ctx    context.Context //nolint:containedctx // Deadline state, not a request scope.
+	cancel context.CancelCauseFunc
+
 	deadline time.Time
 	state    deadlineState
 	pending  uint8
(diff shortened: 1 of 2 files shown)


deadline/deadline.go
```diff
@@ -74,7 +108,7 @@ func (d *Deadline) Set(setTo time.Time) {
 		d.pending--
 		d.state = deadlineStopped
```
noboruma: Nit here, could we rename this `deadlineSuspended`?
I feel `Stop` could be misinterpreted as `cancel`, whereas `suspended` implies the context returned by `Context()` is still alive

noboruma (approved):
LGTM - actually makes good sense to have this `Context()` function out of the deadline.
I will rebase #396 as soon as this PR is merged

deadline/deadline.go
```diff
@@ -47,15 +50,46 @@ func (d *Deadline) timeout() {
...
+
+	return cancel
+}
+
+// Context returns a context for the current deadline, canceled with
+// context.DeadlineExceeded as its Cause.
+func (d *Deadline) Context() context.Context {
+	d.mu.Lock()
+	defer d.mu.Unlock()
+	if d.ctx == nil {
```
JoTurk, line 91: Does this mean we'll return a new context if the context cancelled by a deadline?

JoTurk: I believe this can break SCTP if we migrate to this API. If stream's Write calls `Context` after the deadline has already expired, it will create a new non-canceled context. If the SCTP write is currently blocked it will then then potentially block forever....

JoTurk (changes requested):

deadline/deadline.go
JoTurk: I think it will be a bigger problem for DTLS if we merge @noboruma PR's https://github.com/pion/dtls/pull/1096/ because we'll have multiple `Context` calls and the deadline can expire between `handshake` for example and `contextWithClose` then we'll just get an unbounded timeout. unless we create a single context and store it in conn but then `Set` will not work (we can make Conn reset the context everytime it calls Set but that will be an anti pattern imo).

JoTurk: I think we can just check for `if d.state == deadlineExceeded {` and return a cancelled context.

paulwe: Good catch, and you're right about both consequences. Fixed in b2a362c.

`Context()` now short-circuits on the state, as you suggested:

```go
if d.state == deadlineExceeded {
    return exceededContext
}
```

`exceededContext` is a single package-level context canceled with `context.DeadlineExceeded` at init. Cancellation is immutable, so one instance is safe to share with every caller and costs no allocation on a path that only runs after a deadline has already blown.

Three regression tests cover it, and all three hang for the full assertion timeout against the old code, which is exactly the symptom you described:

- `ExceededReturnsCanceledContext` -- `Set(past)` then `Context()`
- `ExceededByTimerReturnsCanceledContext` -- same via the timer
- `DerivedFromExceededIsCanceled` -- `context.WithCancel(d.Context())` is canceled immediately, with the cause propagating through as `DeadlineExceeded`

`SuspendAfterExpiryGivesLiveContext` pins the other direction: once `Set` re-arms or suspends, `Context()` goes back to handing out a live one.

deadline/deadline.go
paulwe: Done in b2a362c -- renamed to `deadlineSuspended`, and the matching test is now `SuspendKeepsContextLive`.

Agreed on the reasoning: with `Context()` in the picture the distinction actually matters, since suspending leaves the context live while expiry cancels it. `SuspendAfterExpiryGivesLiveContext` pins that.

deadline/deadline.go
JoTurk: thank you for the quick fix

JoTurk (approved):

PR #398, opened 2026-09-09
Drop context.Context implementation from Deadline

paulwe:
Stacked on #397 — review that one first; this PR's diff is the single commit on top.

`Deadline` never satisfied the `Context` contract, for the reasons in #397: `Set` revives it, so `Err()` goes from non-nil back to nil and `Done()` returns a new channel while earlier holders still see the old one closed, and any context derived from it can panic inside `context` with `missing cancel error`.

`var _ context.Context = (*Deadline)(nil)` asserts a contract the type provably breaks. Now that #397 offers `Deadline.Context()` as the honest alternative, this drops the claim: deleting the no-op `Value` method breaks the interface, which turns every derivation from a `Deadline` into a compile error instead of a runtime panic.

`Set`, `Done`, `Err` and `Deadline` are all kept — across `pion/transport`, `pion/dtls` and `pion/sctp` the only methods ever invoked on a `Deadline` are `Set` and `Done`. Nothing in any of the three exposes a `*deadline.Deadline` in exported API; every holder is an unexported struct field.

### Blast radius

Measured by building all three repos against this branch with `-gcflags=-e`, test packages included. Exactly two call sites break, both passing a `Deadline` to an unexported function that takes a `context.Context`:

| repo | site |
|---|---|
| pion/dtls | `conn.go:586` — `c.contextWithClose(c.writeDeadline)` |
| pion/sctp | `stream.go:330` — `sendPayloadData(s.writeDeadline, chunks)` |

`pion/transport` itself needs no changes: builds, vets and passes all tests.

Both are mechanical, and both sites are already coping with the unsoundness:

- `contextWithClose` (dtls `conn.go:858`) does `context.WithCancelCause(context.WithoutCancel(ctx))` and then hand-rolls the propagation in a goroutine with `err := ctx.Err(); if err == nil { err = context.DeadlineExceeded }` — that nil guard *is* the non-monotonicity, already observed and patched at the symptom. It can take `c.writeDeadline.Context()` and drop both the `WithoutCancel` and the guard, or become its sibling `contextWithCloseAndWriteDeadline`, which already watches `c.writeDeadline.Done()` as a plain channel.
- sctp's `sendPayloadData` does `case <-ctx.Done(): return ctx.Err()`. With a revived `Deadline` that returns **nil**, so an aborted write currently reports success — a silent wrong result on `master` today, which this change converts into a compile error.

### Note on versioning

Removing a method from an exported type is formally breaking for `transport/v4`, so this may want a major bump — or the view that a type which never satisfied the contract was never legitimately a `Context`. Flagging for maintainers rather than assuming.


noboruma (approved):

deadline/deadline.go
```diff
@@ -157,8 +154,3 @@ func (d *Deadline) Deadline() (time.Time, bool) {
 
 	return d.deadline, true
 }
-
-// Value returns nil.
-func (d *Deadline) Value(any) any {
-	return nil
-}
```
JoTurk, line 177: we should do a major release when we merge this.

JoTurk (approved):
Thank you for the quick fix.

deadline/deadline.go
paulwe: Agreed -- removing a method from an exported type is breaking, so this wants a major bump. Flagged in the PR body too; happy to leave the sequencing to you.

Worth noting the two downstream call sites this turns into compile errors still need fixing before it can land: `pion/dtls conn.go:586` and `pion/sctp stream.go:330`. Both become `.Context()` calls once #397 is in.

JoTurk: I think we can get more changes to pion/transport and release transport@v5 too. There are other stuff that we can improve.

paulwe: sgtm. do we want to hold on landing this one?

JoTurk: I think yeah, so other people don't accidentally tag it in v4 by mistake

PR #406, opened 2026-09-22
Add background network monitoring and stdnet access

JoTurk:
#### Description
Add `Start` and `Stop` to detector for background monitoring. using a background watcher loop that's built into the background polling or the platform events.
Adds `WithPollInterval` for caller to control the polling and `WithFallbackPolling` for callers to set fallback Polling on native platform in case the network change event got dropped (pretty common).

Adds `GetNet` to get the cached stdnet.Net with its interfaces.

Also made Detector copy friendly.


netchange/detector.go
```diff
@@ -77,6 +117,126 @@ func NewDetector(options ...Option) (*Detector, error) {
...
+func WithFallbackPolling(interval time.Duration) Option {
+	return func(detector *Detector) {
+		detector.fallbackInterval = interval
+	}
+}
+
+// GetNet returns the latest network snapshot, including after Stop or Close.
+func (d *Detector) GetNet() stdnet.Net {
+	return *d.network.Load()
+}
```
JoTurk, line 139: Should we return a pointer?

Sean-Der (approved):

PR #407, opened 2026-09-22
Add global stdnet cache

JoTurk:
#### Description
https://github.com/pion/ice/pull/1003


stdnet/net.go
```diff
@@ -21,10 +23,35 @@ const (
...
+		n, err := NewNet()
+		if err != nil {
+			return nil, err
+		}
+		if !defaultNet.CompareAndSwap(nil, n) {
+			continue
+		}
+
+		time.AfterFunc(5*time.Second, func() {
+			defaultNet.CompareAndSwap(n, nil)
```
Sean-Der: Why `CompareAndSwap` why not just Store?

stdnet/net.go
```diff
@@ -38,6 +65,13 @@ var _ transport.Net = &Net{}
 // UpdateInterfaces updates the internal list of network interfaces
 // and associated addresses.
 func (n *Net) UpdateInterfaces() error {
+	if n == nil {
+		defaultNet.Store(nil)
+		_, err := getDefaultNet()
```
Sean-Der: Should this set `n` ?

stdnet/net.go
JoTurk: to make it UpdateInterfaces / update safe if we update it in the future from ice (originally it was updated with detect and i kept it when i trimmed detect). I don't think there is a big downside for using compareAndSwap.

stdnet/net.go
JoTurk: no? because we just refresh the default / shared value.

stdnet/net.go
```diff
@@ -21,10 +23,35 @@ const (
 
 // Net is an implementation of the net.Net interface
 // based on functions of the standard net package.
+// A nil *Net uses a shared interface cache that expires after 5 seconds.
 type Net struct {
 	interfaces []*transport.Interface
 }
 
+var defaultNet atomic.Pointer[Net] //nolint:gochecknoglobals
```
Sean-Der: what about `defaultNet = &net{}`.

Then do atomic.Value for `interfaces`

Code below then can just do `n = defaultNet` becomes a lot simpler!

JoTurk: I went with a snapshot design because as far as i understand ice calls interfaces in many places, and it doesn't seem safe to change the interfaces during gather, I think it will break mdns

this is what i meant by this comment yesterday https://discord.com/channels/1352636971591274548/1359189787768258671/1551019660214927442

Sean-Der: Do a atomic swap on interfaces. It should be thread safe!

Sean-Der: Just trying to reduce the code as much as possible.

JoTurk: the concerns is above are about consistency / logic safety, not thread safety, we'll update the default stdnet before every gather phase, but if interface changes during a pass we can have unexpected behavior because we call interface 3/4 times.
The other option is to make gather call interfaces once i can do this but it's going to be a really big and ugly change code wise

JoTurk: https://github.com/pion/ice/blob/e50946b37e6f3f95cfd0ccf47578f3fc654675b8/gather.go#L318

and then

https://github.com/pion/ice/blob/e50946b37e6f3f95cfd0ccf47578f3fc654675b8/gather.go#L560

can create broken candidates that are impossible to debug, if interfaces changes between these two

we can make agent call interfaces once per gather, but as i said this can be an ugly change or we'll have to cache interfaces in Agent

JoTurk: @Sean-Der i dropped the snapshot and made it simpler but we'll have to merge https://github.com/pion/ice/pull/1005 first, I didn't go with making Net hold atomic.Value because this will break the copy contract.

JoTurk:
not super safe.

